PR #184, opened 2016-09-28
Feature/178 slim down backstop json

### LINKED ISSUE #178: Slim down backstop.json
We have a lot of redundant options in there. Looks like they could be handled with defaults instead. https://github.com/garris/BackstopJS/issues/127

sonniesedge:
# Summary
Fixes #178.

# Changes

The following changes are proposed in this pull request:
- Remove redundant/empty entries in the `backstop.json` file. 
- Move some repeated entries to the top-level of `backstop.json`, where they can be accessed as defaults by Backstop.

### DIFF
--- a/backstop.json
+++ b/backstop.json
@@ -22,2087 +22,1363 @@
     {
       "label": "Button, generic - default",
       "url": "http://localhost:3000/section/1.1.1-1/fullscreen",
-      "hideSelectors": [],
-      "removeSelectors": [],
       "selectors": [
         ".object-section .o-button"
       ],
-      "readyEvent": "styleguideRendered",
       "delay": 20,
-      "misMatchThreshold" : 0.1,
-      "onBeforeScript": "",
       "onReadyScript": ""
     },
     {
       "label": "Button, generic - hover",
       "url": "http://localhost:3000/section/1.1.1-2/fullscreen",
-      "hideSelectors": [],
-      "removeSelectors": [],
       "selectors": [
         ".object-section .o-button.pseudo-class-hover"
       ],
-      "readyEvent": "styleguideRendered",
       "delay": 20,
-      "misMatchThreshold" : 0.1,
-      "onBeforeScript": "",
       "onReadyScript": ""
     },
     {
       "label": "Button, generic - focus",
       "url": "http://localhost:3000/section/1.1.1-3/fullscreen",
-      "hideSelectors": [],
-      "removeSelectors": [],
       "selectors": [
         ".object-section .o-button.pseudo-class-focus"
       ],
-      "readyEvent": "styleguideRendered",
       "delay": 20,
-      "misMatchThreshold" : 0.1,
-      "onBeforeScript": "",
       "onReadyScript": ""
     },
     {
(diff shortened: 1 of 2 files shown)


backstop.json
```diff
@@ -16,2091 +16,1202 @@
...
-      "hideSelectors": [],
-      "removeSelectors": [],
       "selectors": [
         ".object-section"
       ],
-      "readyEvent": "styleguideRendered",
       "delay": 20,
-      "misMatchThreshold" : 0.1,
-      "onBeforeScript": "",
       "onReadyScript": ""
```
planktonic: Can remove this?

backstop.json
```diff
@@ -16,2091 +16,1202 @@
...
-      "removeSelectors": [],
       "selectors": [
         ".object-section .o-button"
       ],
-      "readyEvent": "styleguideRendered",
-      "delay": 20,
-      "misMatchThreshold" : 0.1,
-      "onBeforeScript": "",
-      "onReadyScript": ""
+      "delay": 20
```
planktonic: I may well have missed the exception(s) as the list is mammoth, but it seems like every scenario uses a `delay: 20` — can we set that globally too? And if there are exceptions can we set it globally then override only for those?

backstop.json
```diff
@@ -16,2091 +16,1202 @@
...
       "selectors": [
         ".object-section"
       ],
-      "readyEvent": "styleguideRendered",
       "delay": 20,
-      "misMatchThreshold": 0.1,
-      "onBeforeScript": "",
-      "onReadyScript": ""
-    },
+      "misMatchThreshold": 0.1},
```
planktonic: There’s a few `mismatchthreshold`s left in here, they can be safely removed?

backstop.json
```diff
@@ -16,2091 +16,1202 @@
...
-      "onBeforeScript": "",
       "onReadyScript": "clickButton"
     },
     {
       "label": "Flex",
       "url": "http://localhost:3000/section/1.10/fullscreen",
-      "hideSelectors": [],
-      "removeSelectors": [],
       "selectors": [
         ".object-section .o-flex"
```
planktonic: Can we also normalise these to all use simply `.object-section`? There are a couple of exceptions — the two modals — which we’d need to be able to override

SonniesEdge:
Okay, things refactored to all have the same selector. Lots of bitmap files have changed, but that's unavoidable, as we've changed so many from selecting an inline element to selecting block elements.

backstop.json
SonniesEdge: This doesn't have a global setting, sadly. Their use of global defaults is a little haphazard.

backstop.json
```diff
@@ -22,160 +22,120 @@
     {
       "label": "Button, generic - default",
       "url": "http://localhost:3000/section/1.1.1-1/fullscreen",
-      "selectors": [
-        ".object-section .o-button"
-      ],
+      "selectors": [ ".object-section" ],
```
planktonic: We can’t remove these in favour of a global setting? :(

PR #197, opened 2016-10-11
Update rails readme info

### LINKED ISSUE #192: Update info on Rails (w/ NPM) integration
The README currently suggests adding the bitstyles node module path via `config/application.rb`. In Rails 4+ however, the asset pipeline configuration has been moved to an initializer of its own:

```
config/initializers/assets.rb
```

which by default contains a separate, commented section:

```ruby
# Add additional assets to the asset load path
# Rails.application.config.assets.paths << Emoji.images_path
```

so maybe you should suggest adding it there as:

```ruby
Rails.application.config.assets.paths << Rails.root.join("node_modules", "bitstyles", "bitstyles")
```

Note: The currently suggested snippet will work, putting it in the initializer might be slightly cleaner though 😃

planktonic:
# Summary
Fixes #192.

# Changes

The following changes are proposed in this pull request:
- Updates the README sections detailing adding bitstyles to your Rails asset pipeline with more correct info regarding Rails 4

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Github release (not from master), and you will need read access to the repo.
 
 ### Bower
 ```shell
-bower install git@github.com:bitcrowd/bitstyles.git#v0.2.2-alpha --save
+bower install git@github.com:bitcrowd/bitstyles.git#v0.7.4 --save
 ```
 
 Once installed, you need to provide your sass installation with the path for bitstyles’ sass. This should be:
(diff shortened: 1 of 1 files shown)


planktonic:
Yes those two sections share a lot. I want the commands to be copy&pastable into the terminal though, so I see no clear way to avoid this, unless you can see a solution?

Right now the Bower & NPM packages are just linked to the github repo, no publishing involved (until very recently the repo was private). I’ll look into the rubygem — charlie handled that.

TL;DR: no we have no automated process for this, that sounds like an issue to me…

planktonic:
https://github.com/bitcrowd/bitstyles/issues/198

Issue #164, opened 2016-08-17
Add CI tests


sonniesedge:
When this project is open-source and publicly available we should create a [free CI server](https://travis-ci.com/plans) and add linting and other build-time tests. 
- Sass linting
- Javascript linting
- Visual regression tests
- CSS stats

planktonic:
- [] Sass linting
- [ ] Javascript linting
- [ ] Visual regression tests
- [ ] CSS stats

planktonic:
I’ll remove the css stats entry, as that’s not a test it should be run locally and checked in deliberately by devs.

planktonic:
I’ve added `styleguide generation` to the list of items we should be testing on travis (just check the styleguide is correctly generated), as a result of #233.

PR #229, opened 2016-11-21
[#164] add ci tests

planktonic:
### Summary

Partial fix for #164.

- Initial `.travis.yml`, currently just running `npm run lint` and `npm run build`.

### Linting and testing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+cache:
+  directories:
+    - node_modules
+rvm:
+  - 2.2
+install:
+  - gem uninstall bundler
+  - gem install bundler --version '1.12'
+before_script:
+  - nvm install 6.9.1
+  - nvm use 6.9.1
+  - npm install
+script:
+  - npm run lint
+  - npm run build


planktonic:
Partial fix for #164 because running the visual regression tests via travis is not simple! Any advice welcome — the problem is not being able to run the styleguide server on travis. Could be done with a hosted styleguide (would mean deploying branches to a styleguide server at the same time), or ngrok to a local box…

pmeinhardt:
Can you just render out the styleguide-server generated pages/individual components to image files in a separate step beforehand and then run the visual regression tests on those? That's what's more difficult on Travis CI, right?

Running on Travis will likely require Xvfb (virtual framebuffer) or something similar.

planktonic:
Our current setup: the styleguide creates the pages from parsing our Sass & the comments therein. The VRT app visits each styleguide URL we specify, uses phantom to render screenshots, then compares them to the checked-in originals.

The hard part is running the styleguide at an accessible URL in order that the VRT can use phantom to render it to images (by all accounts phantom is not a big problem on travis). Also, how do we to specify URLs in such a way as it works both locally and via travis without maintaining separate versions?

Perhaps we chat about this when you have time?

pmeinhardt:
Yup, let's stick our heads together to figure out a solution 👍

planktonic:
In order to get this first step of travis configuration on master I’ll merge this now, and revisit the styleguide question in a new PR

PR #244, opened 2017-01-04
[#231] add parameters to the fixed ratio mixin

### LINKED ISSUE #231: Add parameters to the fixed-ratio mixin
So we can specify which ratio(s) we want. Possibly add a mixin for outputting variants (without respecifying all the properties of the base class). Keep defaults at 16:9 for the base mixin.

planktonic:
Fixes #231 

- Made the `fixed-ratio` mixin a little more general, now accepts width & height parameters to set its ratio.
- Added a new mixin to generate variations without re-specifying all the base declarations.
- Factored out the calculation of ratio into a function.

### Linting and testing

### DIFF
--- a/scss/bitstyles/objects/_fixed-ratio.scss
+++ b/scss/bitstyles/objects/_fixed-ratio.scss
@@ -2,9 +2,9 @@
 //
 // Fixed-ratio
 //
-// Enforces an aspect ratio on this element and its contents. Default ratio is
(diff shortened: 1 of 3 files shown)


planktonic:
Good suggestion 👍 Default ratio is now specified using a variable (I kept the base class still outputting a default size without modifier class as most projects only use one ratio object).

PR #246, opened 2017-01-04
[#227] revisit breakpoints

### LINKED ISSUE #227: Revisit breakpoints
The names of our breakpoints can be confusing. As we no longer use the `@` (e.g. `.o-block@medium-and-up`) to signify 'breakpoint!' their names should make it somehow clear that they refer to viewport size/breakpoint.

It’s also worth looking at which breakpoints we really need — we have many defined as default in the breakpoints list.

And it maybe useful to define some presets, even if they’re not then used in the breakpoints list e.g. landscape/portrait.

planktonic:
Fixes #227 

- All media queries are now mobile-first (i.e. target all viewports at the specified size *and up*), except for that targetting smallscreens only.
- The smallscreen media query has been renamed `small-only` to make that clear, and distinct from the others
- Replaced the `-`-delimited naming for classes at breakpoints with `@`-delimited, for consistency (existed on `hidden` & `visuallyhidden`) and clarity (clear it’s special, not just part of the classname)
- Added the landscape & portrait queries

- [x] `npm run test:build` script has been sucessfully run and new/changed images committed

### Linting and testing


RRorg:
fixed typo, everything alrighty now

scss/bitstyles/layout/_width.scss
```diff
@@ -11,10 +11,10 @@
 // `.l-width--1-of-12` … `.l-width--12-of-12`.
 //
 // If you define `$bitstyles-widths-breakpoints` to contain the names of one or more
-// breakpoints (defaults to just `medium-and-up`), widths will also be created
+// breakpoints (defaults to just `medium`), widths will also be created
 // for those breakpoints:
 //
-// `.l-width--4-of-12-medium-and-up`
+// `.l-width--4-of-12-medium`
```
planktonic: I’d say just leave a line comment when making your review — the PR author can fix it/suggest something else and update their PR.

But in this case yup that’s fine :)

RRorg: I think you're not notified when I leave line comments, thus them going unanswered - I'll remember to post a "real" comment referring to them next time :)

planktonic: Hmm, I get email notifications for line comments (at least when part of a review) — I got an email about your original question and your last comment. Perhaps your settings are different?

PR #253, opened 2017-01-05
[#248] unit management

### LINKED ISSUE #248: unit management
We currently have only one global unit of measurement: `$bitstyles-spacing`. Using a map to provide a scale would be more useful for client projects.

e.g.
```scss
$bitstyles-sizes: (
  small: xx,
  base: xx,
  large: xx
);
```

Suggestion: default to using `base` size everywhere we currently use `$bitstyles-spacing`. Also, it should default to being equal to base font-size (also changing based on media query?).

planktonic:
Fixes #248 

- Replaced the `$bitstyles-spacing` variable with a map of variables to allow several sizes (most often used for margins & padding) while keeping consistency.
- Added an accessor function to aid readability over the sass `map-get($map-name, 'map-item')` function call.

### Linting and testing


planktonic:
Rebased & updated styleguide docs

RRorg:
I feel like `size` is a bit ambiguous, can we use `spacing` instead?
Maybe change `$bitstyles-sizes` to `$bitstyles-spacing-sizes` too?

planktonic:
Good call. Updated

PR #254, opened 2017-01-05
[#252] colour helper functions

### LINKED ISSUE #252: Colour helper functions
Currently a map is used for management of colours. Each time they’re used with e.g. `map-get($bitstyles-colors-theme, 'brand')`, which could be more readable & less error-prone if there were helper functions e.g.

```scss
palette('brand');
```

Suggestions:
- rename `$bitstyles-colors-theme` to `$bitstyles-palette`
- add light & dark variations e.g.
  ```scss
  palette('brand', 'dark')
  ```
  (default to 'base')

(Breakpoints are also a map, but as breakpoint usage is already abstracted into a mixin and never used directly, a helper function seems unnecessary there)

planktonic:
Fixes #252

- Renamed `$bitstyles-colors-theme` to `$bitstyles-palette`.
- Added capability for colour tones in the palette:
    - `base`, `light` and `dark` tones specified for each of the colours in the palette (only `base` is required for a colour)
- Created `palette` as an accessor function for the colour palette
- Replaced direct uses of `map-get($bitstyles-colors-theme, 'color')` with an accessor call.

### Linting and testing


RRorg:
Should I squash and merge when I'm the final reviewer to approve?

planktonic:
@RRorg No it’s fine, I can sort that out thanks

PR #256, opened 2017-01-09
[#251] typography management

### LINKED ISSUE #251: typography management
Right now there’s a series of variables defining font-sizes for each `<hx>` element and body font on smallscreens. Another variable (`$bitstyles-typography-breakpoint`) dictates at which breakpoint these font-sizes are redefined for the large versions of these variables. This is rigid and limiting: there must be two and only two sets of font-sizes, and all sizes must be defined in both sets.

This may be better implemented using a map — less boilerplate code, fewer variables, and a more flexible system allowing multiple breakpoints e.g.

```scss
$font-sizes: (
  'small': (
    'html': xx,
    'h1': xx,
    'h2': xx,
    'h3': xx,
    'h4': xx,
    'h5': xx,
    'h6': xx
  ),
  'medium': (
    'html': xx,
    'h1': xx,
    'h6': xx
  )
);
```

Then using a mixin to iterate over this map, outputting font-sizes for each element defined, at each breakpoint defined (breakpoints need to exist in the global breakpoint map as with all other maps).

Notes:
- I’m not sure how best to provide a 'base' set of font-sizes in a mobile-first manner within this system — the mixin outputting font-sizes needs to output the first set of sizes without a media-query.
- It’d be great not to *have* to redefine each font-size at each breakpoint — as the breakpoints are all soon to be `min-width`-based, the previously-set font-size will be used.

planktonic:
Fixes #251 

- Replaced the multiple variables specifying font-sizes with a single map.
- Added mixins & functions to use that map.
- Removed the uneccessary duplication of font-family when the specified webfont is the same as the base font (avoids increased specificity unless necessary).
- Fixed a bug whereby base font-size at larger screensizes was incorrectly set.
- Bugfix resulted in changed appearance for many elements at larger screensizes — PR also updates reference images for those elements/objects

- [x] `npm run test:build` script has been sucessfully run and new/changed images committed

### Linting and testing


RRorg:
Could you re-check the tests? 63 failing for me

planktonic:
I’ve rebased to master.

@RRorg all tests pass for me, which fail for you?

planktonic:
See comment on #258 before merging: https://github.com/bitcrowd/bitstyles/issues/258#issuecomment-272838888

RRorg:
@planktonic working now!

planktonic:
@RRorg great!

As per comments in #258 I’ve updated the styleguide config when running for tests:

```npm run styleguide:test```

This disables the anti-aliasing and should fix issues with having font-smoothing enabled/disabled on macOS (please check that!). Could also fix similar issues between platforms (windows/linux) but I haven't been able to test that.

Sadly this change means we’re generating a 'special' version of the styleguide just for testing purposes — ideally we test the live system, not some version altered to pass tests — but I see no way around that except to make the styleguide useless as an example of bitstyles’ output.

RRorg:
Can you give me 24 hours to see if there is a better way to get around the AA testing issue? I'm not yet willing to accept that this has to be the best way 😞

planktonic:
Yup, please do!

RRorg:
Please hold off on merging this for now, I am currently working on a way to make the BackstopJS font rendering consistent across devices.

RRorg:
the new commit fixes #258 by forcing the anti-aliasing option to be off across all setups and operation systems

it also increases the `misMatchThreshold` to `0.15` from `0.1` to better account for slight OS differences

when creating new Backstop scripts, please remember to `require('./onLoad.js')` and initialize it with the Casper object (see current scripts)

planktonic:
I’ve rebased this to master & made a small fix for background-colour on the styleguide fullscreen pages.

I also regenerated the VRT reference images — @RRorg the images in your last commit wouldn’t validate for me regardless of whether font-smoothing was on or off! Can you pull & run the tests again on your machine?

RRorg:
Yes, this is the good stuff!! Everything working fine now, independent of font smoothing settings!
Full testing satisfaction achieved. 👌

Sorry, I have no idea, something had to have been flakey with my VRT images, although they ran fine on my machine.

Another small improvement I suggest regarding the way test tasks are run: currently we're chaining them with `&&`, which means the tests won't continue when one of them failed. I suggest changing the chaining to use `;`, so all tests run through even if one of them failed and one has a complete overview of what needs fixing.

planktonic:
@RRorg phew! great. I think at one point you mentioned using imageoptim of similar to optimise the images, did you do that with your new images? I recall it giving me problems with image comparison…

If this PR is good to go I’d like to merge this now and look at the test running (`&&` vs `;`) in a separate PR — could you add an issue for that please?

Issue #259, opened 2017-01-12
Add yarn support?


planktonic:
There’ve already been a couple of issues with node package versions — on a fresh install of bitstyles the styleguide and/or visual regression tests fail or error due to e.g. parsing errors introduced with a new version.

Yarn is compatible with npm — uses the same packages and the same `package.json` — but it adds a `yarn.lock` that ensures devs who install with yarn instead of npm will get exactly the same package versions.

planktonic:
Fixed by #270, closing

PR #270, opened 2017-01-19
[#259] Add Yarn Support

RRorg:
A possible solution for #259

- update all Dev-Dependencies to their latest stable releases
- use yarn instead of npm for this project

- [X] `yarn run checks` script has been run without errors.

Implementing the yarn support like this is not a small step, since it *requires* contributors to use yarn. A "softer", more tolerant approach would be merely adding and maintaining the `yarn.lock` file, while still running all scripts via npm, thus making yarn optional.

Forcing contributors to use yarn is a potentially controversial step, so let's only do this if all current contributors are in favor of it. I've marked you as reviewers.

Please delete your `node_modules` folder when switching over from npm, and reinstall with yarn.

ninja edit: I didn't want to mess with the Travis settings, so let's not forget changing them when we switch over.


RRorg:
I feel like Travis support for Yarn is just not yet there, at least, until it comes pre-installed on the build images. Let's just update the dependencies and add an optional Yarn lockfile.

pmeinhardt:
Yarn should be there on Node.js images I think. Looking at the `.travis.yml` it seems like no `language: node_js` is present though… The default image is Ruby if I remember correctly. Adding this to the top of the config may be worth a try?

RRorg:
Good call, I'll revisit this!

README.md
```diff
@@ -190,9 +190,11 @@ Clone this repository, `cd` into the folder created, then install dependencies:
 ```sh
 git clone https://github.com/bitcrowd/bitstyles.git
 cd bitstyles
-npm install
+yarn
 ```
 
+It is recommended to use `yarn` instead of `npm install` to ensure version consistency across dependencies.
```
planktonic: Can you add a link to yarn here please? e.g. https://yarnpkg.com/en/docs/install

package.json
```diff
@@ -9,21 +9,18 @@
     "lint:scss": "postcss -c postcss_lint.json",
     "lint:js": "eslint .",
     "lint": "npm run lint:scss && npm run lint:js",
-
```
planktonic: Was this deliberate or just automatic editor behaviour? I’d prefer to keep this section chunked as it’s so long; if necessary add blank comments.

RRorg: sorry, my bad, has to have been something automatic

package.json
```diff
@@ -60,22 +57,22 @@
...
+    "cssnano": "^3.10.0",
+    "eslint": "^3.13.1",
     "eslint-config-bitcrowd-base": "^8.0.0",
-    "eslint-plugin-import": "^1.16.0",
-    "node-sass": "^3.10.1",
-    "parker": "1.0.0-alpha.0",
-    "postcss": "^5.2.4",
+    "eslint-plugin-import": "^2.2.0",
+    "node-sass": "^4.3.0",
+    "parker": "^0.0.10",
```
RRorg: reverted

RRorg:
If you're all in favor, this can be merged now. Not full-on-yarn yet, but with added yarn support.

image

planktonic:
Thanks, looks good!

PR #273, opened 2017-01-30
[#271] icon sizes should be specified in a map for consistency

### LINKED ISSUE #271: icon-sizes should be specified in a map for consistency
`$bitstyles-icon-sizes` is currently still specified as a Sass list instead of map — but it’s very much a map structure and should be specified as such for consistency.

(Probably a left-over from when bitstyles required a version of sass too old to support maps)

Also check there are no other good candidates for this change.

planktonic:
Fixes #271 

- `$bitstyles-icon-sizes` is now a correct Sass map.
- `@mixin generate-font-sizes` and `@mixin output-font-sizes` both updated to use the more concise Sass map `@each` iterator accessors (instead of one of the more verbose boilerplatey Sass accessor functions like `map-get` or `nth`).

Every time:

### DIFF
--- a/scss/bitstyles/objects/_icon.scss
+++ b/scss/bitstyles/objects/_icon.scss
@@ -35,12 +35,9 @@
   fill: currentColor;
 }
 
-@each $icon-size in $bitstyles-icon-sizes {
-  $alias: nth($icon-size, 1);
-  $size: nth($icon-size, 2);
-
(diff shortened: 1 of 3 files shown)


planktonic:
Travis red due to issue on master. Re-check once that’s fixed & this is rebased.

PR #275, opened 2017-01-30
Make Travis use the NodeJS environment & yarn

RRorg:
hotfix changing the `.travis.yml` file so Travis uses yarn instead of npm and fetches a working version of browserslist


.travis.yml
```diff
@@ -1,15 +1,21 @@
...
     - node_modules
-rvm:
-  - 2.2
+language: node_js
+sudo: false
+before_install:
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  - export PATH=$HOME/.yarn/bin:$PATH
+node_js:
+  - "5.7"
```
planktonic: If we’re specifying a version here, can we remove the `nvm` commands below? (and bump this to v6.9.1 instead)

RRorg: Sure, and let's not forget to update the .nvmrc then too. do you want to handle this or can you wait for me to return home?

planktonic: Ok, I’ll update & merge just so we get this fix on master quickly. Thanks!

Issue #269, opened 2017-01-19
Repo image bloat


RRorg:
Just so we don't forget about it, we need to separate the reference images from the repo someday™

todo: check how other projects are handling this issue

planktonic:
Could be related to #163 and #164

RRorg:
We could remove the reference images from the repo completely, and instead publish a precompiled version to the #gh-pages branch. Then, on the VRT step, generate fresh reference images from bitcrowd.github.io/bitstyles/ and compare them to the images created from the local project. Reference files could be locally cached too and regenerated on command/every X time units. That should kill both the bloat as well as the inconsistencies between dev machines.

planktonic:
Fixed by #277

Issue #163, opened 2016-08-17
Hosted version of styleguide


sonniesedge:
If and when Bitstyles is made publicly available, we will need a hosted version of the styleguide that can be linked to from the documentation. 

The styleguide will serve to demo the usefulness of the library and possibly provide a stable reference for visual regression testing.

planktonic:
I’ll close this issue finally, as for a while we’ve had a hosted version of the styleguide available at https://bitcrowd.github.io/bitstyles/ 🙏

PR #277, opened 2017-01-31
[#164] add ci tests

### LINKED ISSUE #164: Add CI tests

planktonic:
Fixes #164 

**WIP do not merge**

- All VRT reference images have been regenerated on an Ubuntu 12.04 Vagrant box
- Visual regression tests now run (& pass!) on Travis
- Added provisioned Vagrant config to automate setup of local environment

There are undoubtedly improvements that could be made here, this is something of a proof of concept. I started from the premise that the phantom tests need to run on travis, therefore copying the environment in a vagrant box seemed logical.

This does add overhead:
- running the VRTs now needs to be done inside a vagrant box (which needs some initial setup, most of which is automated)
- at some point we’ll likely need to update the environment of the vagrant box (because npm packages used in bitstyles changed, or to mimic Travis’ environment). This will mean updating the vagrant config and recreating your local box, or manually updating the environment inside the already existing box.
- if you’re not deving bitstyles inside the vagrant box (which I don’t expect to be a great experience), you’ll have to push from outside vagrant, then once inside pull & switch branches before testing

Can we mitigate the extra complexity all this adds? Is it a price worth paying to get the VRTs running on travis?
Do we need the font-smoothing fix we added last week? (it was for compatibility between macs when font-smoothing preferences are set differently)

Delete if not applicable:

- [x] Changed the visual appearance of object(s), or changed the tests? `npm run test:build` script has been successfully run and new/changed images checked visually for correctness and committed.

Every time:


planktonic:
@pmeinhardt @RRorg feedback on this one greatly appreciated! I’m basically new to vagrant so it’s very possible this could be done better (perhaps even without vagrant, actually?)

RRorg:
The workflow feels very clumsy, with duplicating the repository. vagrant could directly mount the project directoy, so you wouldn't need to ssh into the box to change branches, pull, etc. But there is no way to use an alternative `node_modules` path (that I know of, maybe it could be overwritten by some hacky symlink magic inside vagrant?), so there's no way I see making the build process work for the same physical project instance in both host and vm, due to npm binary compiling.

That being said, and without being able to offer a better solution, I just want to emphasize that vagrant comes with a heavy price tag:
- a whole VM system with an external setup to complete, which also has to be maintained by us
- disk space consumption, download time (the precise64 box download is by now taking over 50 minutes for me), memory consumption
- hardware and OS limits, vagrant depends on hardware side virtualization (VT-X, AMD-X, etc) - this will unnecessarily lock out developers on older/cheaper devices or more exotic OS configurations (Hackintosh, some Linux distributions)

We'd be adding quite a few extra layers of bloat, complexion, maintenance and gateway hurdles for what basically amounts to comparing a few bitmaps. At that point, if I had to make a call I'd rather scrap VRT completely 😐

pmeinhardt:
I've last used Vagrant for virtualization about two to three years ago. Back then I think it was a great tool… (you know where this is going 😁)

For what you're trying to achieve, I'd go down a different road, *if visual regression tests are a must for you even though apparently, you'll get different results based on platform/installed fonts/…*.

**Since you're running these tests on Travis, it would make sense to me to make the render output generated in that environment the "source of truth".**

To make testing consistent, I'd use the Docker image used in the Travis CI tests and just run the visual regression tests in a container based on their image, in your case: https://quay.io/repository/travisci/ci-nodejs.

The [`Dockerfile`](https://docs.docker.com/engine/reference/builder/) would look something like this:

```Dockerfile
FROM travisci/ci-nodejs

USER travis

RUN mkdir -p ~/bitstyles
WORKDIR ~/bitstyles

# create a layer with the dependencies
COPY package.json package.json
# or yarn install, whatever
RUN npm install

COPY . ~/bitstyles

CMD ["npm", "run", "styleguide"]
```

The [`.dockerignore`](https://docs.docker.com/engine/reference/builder/#/dockerignore-file) would look almost exactly like the `.gitignore`, but at the very least, it would ignore the `node_modules`.

Steps for testing would then be (could be provided as a script):

1. `docker pull quay.io/travisci/ci-nodejs`
2. `docker build --tag bitstyles-ci .` (in the project's root directory, where you place your `Dockerfile`)
3. `docker run --rm bitstyles-ci npm run styleguide`

On Mac and Windows you'll need to use `docker-machine` and set up the environment before: `eval $(docker-machine env)`.

To inspect the generated output images, you could mount a directory on your machine as a volume for the container. Maybe you should do that anyway, if you plan on committing the updated images for wanted changes.

Hope that helps?

planktonic:
Lots of great feedback, thanks so much both! 👍 

Initial thoughts (I’ll digest and maybe come back with more later :D), correct me if I’ve misunderstood anything:
- VRTs are always dependent on browser/fonts/platform (even with the same rendering engine, checkboxes etc. look different). The thing we’re looking for is *change*. That’s fine. Some of our current reference images don’t actually match what you see in browser (because phantom); what’s important is that they don’t unexpectedly change.
- I’m thinking this morning that it’s possible the font is the main issue. Our original reference images used helvetica because we’re on Macs, this doesn’t exist as standard on linux. If we can avoid any kind of virtualisation by changing our default font to one that exists on both linux & macs, that may be a way forward — I’ll look into that this morning.
- As you both seem to agree vagrant isn't the way forward(?), I’ll look into the Docker workflow if the font change seems unusable for whatever reason.
- If this PR doesn’t go anywhere, that’s totally fine! It’s exploratory. In case of failure I’d update the corresponding issue with what I’ve found here, and remove the requirement for running the VRTs on Travis (that’s the aim here, the virtualisation was just a way to get the same rendering locally & on travis).

RRorg:
@planktonic @pmeinhardt thoughts about this? https://github.com/bitcrowd/bitstyles/issues/269#issuecomment-276466315

RRorg:
I mean basically what we need is a reference that has the "before" state so we can compare it with the "after" state. Nobody says the before state has to be in the form of prerendered images, and nobody said the "before" has to look exactly the same on all devices... 😋

planktonic:
Yes true, if we could render the reference images from master on travis for each PR, then checkout the current branch and run the tests (render & compare), that may solve the issue of platform incompatibility.

Though that would approximately double the time it takes the tests to complete.

We’d also loose the ability to run the tests locally, unless we also provide a docker-based solution as above (or something similar).

planktonic:
Ah I see, *EDIT:* no need to checkout master if we have hosted styleguide

planktonic:
Ok, so based on discussions with both @pmeinhardt and @RRorg today, we have a few potential solutions. These are my notes on what we’ve discussed, please let me know if i missed/misunderstood anything.

We could **keep committing reference images & use a docker virtualisation process** as paul outlined above. This results in tests always being run in the same environment, so no rendering differences. It adds some complications (install & setup docker, scripts to manage the process) & a dependency in the form of Docker.

The remaining two options *require reference images to be regenerated on each feature branch*, as they would no longer be committed to the repo. So test time increases but we remove generated files from the repo (see #269).

**GitHub Pages** version:
  - We’d have a hosted styleguide on GH pages (see #163) 👍
  - Adds dependency on an external resource & the network 👎

  Process:
  - Initially, on master:
      - generate styleguide
      - commit generated html of styleguide
      - Set as gh-pages content so the styleguide is available at `bitcrowd.github.io/bitstyles/`
  - On feature-branches & on travis:
      - generate reference images from `bitcrowd.github.io/bitstyles/` styleguide
      - generate local styleguide from feature-branch
      - generate test images, compare against those generated from gh-pages
      - commit styleguide changes


**Local subdirectory** version:
  - Self-contained, local 👍
  - No hosted styleguide (at least not as a direct result of this; doesn’t preclude the GH pages thing) 👎
	
Process locally & on travis:
 - clone repo from GH into a sub directory
 - checkout master
 - generate styleguide
 - generate reference images
 - copy those reference images to 'real' local version of repo (backstopjs config is the same between  master & feature-branch, it’s complicated by subdirectory)
 - back up to local branch
 - generate styleguide
 - generate test images, compare against those generated above


Of course for both of the these the steps would be scripted so it’d just be one command.

As ever, feedback welcome!

planktonic:
To try it out, I’ve updated this branch with the 'local subdirectory' version. Tests are currently passing on Travis and locally (I’m re-running tests right now to check there’s no flakiness).

It works well; my biggest issue with this approach is the time it takes  — local test runtime is up from 4minutes to 10mins. This is due mainly to regenerating the references images every time (and checking out & `npm install`ing a new copy of the bitstyles repo takes a couple of minutes).

@RRorg can you try this out locally? Check it works for you, but also please measure the runtime.

RRorg:
It does work for me and the test finishes in 9 minutes (MacBook Pro Mid 2013)

Good to finally have a solution that is both automatic and cross-platform-compatible, even if it's slow and kinda, somewhat, slightly ugly 😉

pmeinhardt:
Haven't had time to look at the implementation details…

Quick thought: To make the "takes twice the time" issue a little less severe, you could cache the reference images by commit id (e.g. `master` is at `3957da62b9e7a2dfd5f27dc1c3b07dda52411380`) and only re-generate, if it has changed.

planktonic:
@pmeinhardt Yes, that’d be great! I’m not sure how to achieve this; can we discuss it sometime?

pmeinhardt:
@planktonic: sure.

planktonic:
TODO if this branch is to be merged:
- [ ] remove the `font-smoothing` fix from this branch
- [ ] remove the `font-smoothing` fix from master

planktonic:
After discussion with @pmeinhardt, we’re now regenerating the reference images only if the commit hash of origin/master has changed since the last time the reference images were generated. This has no effect on the first run (nor test runs after master has been updated), but it reduces the runtime of subsequent test runs back to ~4m (locally).

The caching also works on Travis, cutting the runtime by half, down to ~6m. I can’t check this thoroughly until/if this branch is on master, but I hope this should be true across all branches we push i.e. if master hasn't changed since any branch generated the images, we can skip regenerating the images on the current branch.

@RRorg @pm if/when you have some time to take a look (I’m back on this monday), any suggestions for improvements are very welcome! Better error handling, code structure… I’m no shell scripter so I imagine there’s plenty of room for improvement here :D

RRorg:
Hmm, I really don't like that we have to rely on taskkill for this. Having a simple testing script for a Sass library go around and kill tasks on my computer - by name! - feels like a big no-no to me. We can maybe do something hacky like using `screen` (does it come preinstalled on Mac?) to start the styleguide process, remember the screen ID and then just abort it instead of taskkilling.

As it is, we could even think about using the github pages version - the downside to this was network & url dependency, but we have this with the current approach too.

planktonic:
@RRorg thanks, I’ve updated this branch again — instead of killing the styleguide by grepping its name out of `ps`, it’s now killing just its own child processes (note that this is complicated by the fact different OSes spawn the new processes here differently; we cannot rely on a consistent structure of process groups nor of parent/child processes). `screen` isn’t on macs as standard.

Network access with this approach is limited to just the request for the github repo, my assumption here is that the process is less exposed to networking issues than if every test case is a new network request.

If you’ve thoughts on that, or suggestions on the code here, please fire away :) (and @pmeinhardt!)

planktonic:
fixes #269

PR #280, opened 2017-02-08
Test styleguide build

planktonic:
Partial fix for #164 

- Added new script to simply build the styleguide (no watch task, no webserver)
- Added `styleguide:build` to travis config.

Every time:


scripts/styleguide-server.sh
```diff
@@ -0,0 +1,20 @@
...
+  --disableEncapsulation \
+  --title 'bitstyles' \
+  --customColors "styleguide/assets/stylesheets/styleguide-colors.css" \
+  --extraHead '<script src="/assets/javascripts/ofi.js"></script>' \
+  --extraHead '<script src="/assets/javascripts/a11y-dialog.js"></script>' \
+  --extraHead '<script src="/assets/javascripts/styleguide-init.js"></script>' \
+  --extraHead '<script src="/assets/javascripts/modal.js"></script>' \
+  --extraHead '<script src="/assets/javascripts/expander.js"></script>' \
+  --extraHead '<script src="/assets/javascripts/object-fit-polyfill.js"></script>' \
+  --extraHead '<link rel="stylesheet" type="text/css" href="/assets/stylesheets/styleguide-theme.css">' \
```
pmeinhardt: The `styleguide-server` script seems to duplicate all of the command-line arguments of `styleguide-build` and just adds two more: `--watch` and `--server`.

I think this could be dealt with simply by using `-build` as the base and allow it to receive additional commands. Something like the following should work:

```shell
#!/bin/sh
# styleguide-build

styleguide \
  --kss-source 'scss/**/*.scss' \
  … \
  $*
```

```shell
#!/bin/sh
# styleguide-server

styleguide-build --watch --server
```

scripts/styleguide-build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
```
pmeinhardt: If you want to be able to run this from a directory other than the project root, adding this simple one-liner at the top should do the trick:

```shell
cd "$(dirname "$0")/.."
```

scripts/styleguide-build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
```
pmeinhardt: I would probably `chmod a+x scripts/*` and then drop the `.sh` extension. But that's just a matter of taste.

scripts/styleguide-server.sh
pmeinhardt: For reference: https://www.gnu.org/software/bash/manual/bashref.html#Special-Parameters

planktonic:
Thanks for the feedback both! @pmeinhardt I made most of the changes you suggested, only leaving the last `cd "$(dirname "$0")/.."` — these scripts are intended to be run via the npm scripts (there is some setup done there too), location should not be an issue.

PR #286, opened 2017-06-08
Tests select .object-section only

planktonic:
- All tests now select the parent element (`.object-section`) of the element(s) to be tested. This ensures that if the markup of a component is changed, the new elements are all included in the test image. Most tests were already doing this (all new tests have been written this way for some time), but we’d held off changing the remaining legacy tests as the costs of changing so many images didn’t seem worth it (now not an issue as no reference images are included in the repo).

Delete if not applicable:

- [x] Changed tests in any of the `backstop_tests/` test suites.
- [x] Changed the visual appearance of object(s), or changed the tests? `npm run test:build` script has been successfully run and new/changed images checked visually for correctness and committed.


planktonic:
This PR shows an issue with our new way of handling reference images in the visual regression tests — this branch will always fail the CI tests, as the images here do not match master. Of course they *should not* match master, that’s deliberate! I’ve checked each locally(!) but that’s not ideal, an breaks the automation we’re trying to attain…

PR #293, opened 2019-10-27
Fix build (wip)

lwassermann:
I just wanted to update the dependencies. But I can not install the  dependencies on my machine. node-sass refuses to build. Therefore, escape to the CI. Which is red 😞. I managed to get a few more step going, but the global dependencies now include code that will not run on node 6.9. So we have to update everything at once, without test-suite support (because the source of red is unclear: is it a dependency, or is it broken behavior).

Maybe this first step helps the next person 🙂

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
 node_js:
   - "6.9"
 install:
-  - gem uninstall bundler
+  - gem uninstall bundler --executables
   - gem install bundler --version '1.12'
   - yarn
   - npm install -g phantomjs-prebuilt casperjs backstopjs
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,9 @@ cache:
     - test/backstop_data
 language: node_js
 sudo: false
-before_install:
-  - curl -o- -L https://yarnpkg.com/install.sh | bash
-  - export PATH=$HOME/.yarn/bin:$PATH
 node_js:
-  - "6.9"
+  - '6.9'
 install:
-  - gem uninstall bundler --executables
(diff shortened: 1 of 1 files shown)


pmeinhardt:
Have you tried updating the Node.js version @lwassermann? No one should be using 6.x anymore anyway.

— https://nodejs.org/en/about/releases/

lwassermann:
No. Because if I update the node version, I can not judge where a red test comes from: Having inadvertently broken the feature or dependencies that changed behavior.

reneroth:
had a similar issue lately, can you try fixing node-sass to `4.10.0`? `^4.3.0` as it is now loads `4.13.0` which does not play nicely with some node versions

pmeinhardt:
🤔 I've tried updating both Node.js and `node-sass`, but to no avail.

E.g. `parker` fails, a tool which hasn't been updated in more than 3 years. To be honest, to me it looks like we need to do a longer round of updates and even changes, followed by a close look 👀 to make sure everything's still working correctly.

pmeinhardt:
I've made a little bit of progress here: https://github.com/bitcrowd/bitstyles/tree/fix-build-2

But it really does seem like we'll need at least half a day to maybe a day to upgrade all of the outdated dependencies, move away from `parker` if we need it at all and get CI green again.

Migrating over to Circle CI would be nice too.

lwassermann:
> I've made a little bit of progress here: https://github.com/bitcrowd/bitstyles/tree/fix-build-2

Awesome. In that case I'll close this PR and branch.

> FYI: One of the errors we get is from natives, a package that nobody should really be using:

😨

PR #295, opened 2019-11-04
Chore/update dependencies

### LINKED ISSUE #294: Update dependencies
Using this package gives many warnings of security vulnerabilities from NPM. Please update where possible, and remove those packages with no recent-enough release.

planktonic:
Fixes #294 

I’ve marked the largest changes with a ‘⚠️’, not because they are a problem from my PoV, but because are interesting, or represent a a feature removal.

- Update npm package versions where possible
- Updates stylelint config to latest available from bitcrowd/frontend
- Removes all post-css packages. Uses just stylelint for linting, and just node-sass for building. This should not affect any projects using bitstyles, as the bitstyles build chain is only used here.
- Removes various metadata files related to publishing the package on bower, and as a gem. These are no longer useful, nor indeed used.
- Updates the node version in `.nvmrc`
- ⚠️Removes yarn.lock for now, as node-sass will only build when using npm 🤔. We should re-introduce this (instead of package.json, not both), when the issue with building node-sass/node-gyp is resolved
- ⚠️ Removes all the styleguide-related files & setup. The SC-5 project is now abandoned. A newer styleguide/component-library should be found
- ⚠️ Removes all the VRT backstopJS-related content, but retains the reference images for possible future use
- Removes css parker, as it’s now unmaintained, and has several vulnerabilities. replaces it with Analyze-css, and updates the stats.json.
- Updates the readme to reflect all the above

EDIT:
This results in one LOW-level vulnerability remaining, present in `stylelint-config-bitcrowd > stylelint-order > stylelint > micromatch > braces`


planktonic:
I’ve created issues to replace the styleguide & VRT runner: #298, and to investigate the problems installing/building node-sass/node-gyp #296

.eslintrc.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "extends": [
+    "airbnb",
+    "plugin:prettier/recommended"
+  ],
+  "plugins": [
+    "prettier"
+  ],
+  "rules": {
+    "prettier/prettier": "error"
```
lwassermann: This is what `plugin:prettier/recommended` also does. Do we need both?
👉  https://github.com/prettier/eslint-plugin-prettier/blob/master/eslint-plugin-prettier.js#L113

Since we're using this together with airbnb, there is also [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier/blob/master/README.md), which selectively deactivates rules that would conflict with prettiers whitespace choices.

.nvmrc
```diff
@@ -1 +1 @@
-6.9
+11.10.1
```
lwassermann: Is there a particular reason to go with node 11? Since October, 12 is LTS.

.stylelintrc
```diff
@@ -1,3 +1,8 @@
 {
-  "extends": "stylelint-config-bitcrowd"
+  "extends": "stylelint-config-bitcrowd",
+  "rules": {
+    "at-rule-no-unknown": [ true, {
+      "ignoreAtRules": ["else", "include", "mixin", "if", "each", "function", "return", "error", "for", "warn"]
```
lwassermann: If we extend our own config in our own lib, should we add this rule to the bitcrowd style config?

README.md
```diff
@@ -213,72 +150,4 @@ npm run build
...
-```
-
-Then recreate the reference images:
-
-```sh
-npm run test:build
-```
-
-Check the new reference images match what you expect, then commit them and you’re done.
+Previously this project used a custom-made visual regression testing system. The dependencies used have sadly become abandoned in the years since, so they have been removed (encouraged by the Github security auditing feature that went live).
```
lwassermann: Maybe add a reference to the issue that adds it back, in case somebody wants to track the progress?

.nvmrc
planktonic: Nope, that was the version I could get node-sass to compile with no problems! I just updated this to use 12.13.0, and node-was compiles for me fine. Could you delete your node_modules and try to `yarn install` again please?

.eslintrc.json
planktonic: No, I think not, good catch! 👍

.stylelintrc
planktonic: Possibly 🤔 We generally only need the `include` keyword in other projects (the other stuff is pretty low-level, and only gets used here), but there’s no reason not just add them all tbh — it’s all the sass keywords. I believe there’s also a plugin to add linting all sass keywords.

For now, I’d suggest we leave this here (won’t pass tests otherwise!), make the update over in the stylelint config, then remove this after, wdyt?

README.md
planktonic: 👍

planktonic:
> I noticed that the README still includes a sentence pointing to yarn. Since we change the lock-file, should we also take out that reference?

@lwassermann if you can successfully run `yarn install` (this works for me with no problems with node v13), I’d say we stick with yarn, and remove the package-lock.json (that’s the current state since my last updates)

lwassermann:
@planktonic Installing node-sass with yarn was successful.

.stylelintrc
lwassermann: Sure. It was just a thought when seeing the file 🙂

README.md
```diff
@@ -150,4 +150,4 @@ npm run build
 
 ## Testing
 
-Previously this project used a custom-made visual regression testing system. The dependencies used have sadly become abandoned in the years since, so they have been removed (encouraged by the Github security auditing feature that went live).
+Previously this project used a custom-made visual regression testing system. The dependencies used have sadly become abandoned in the years since, so they have been removed (encouraged by the Github security auditing feature that went live). there is an issue open to find a suitable replacement: [https://github.com/bitcrowd/bitstyles/issues/298](https://github.com/bitcrowd/bitstyles/issues/298)
```
lwassermann: `There` (beginning of the new sentence).

.eslintignore
```diff
@@ -1,5 +1,4 @@
 node_modules
 build
-styleguide
```
lwassermann: 💖

lwassermann:
👍 ⛵️

README.md
lwassermann: (Also maybe shorten the link label to `#298`?

.stylelintrc
planktonic: Yup, thanks! I added an issue here: https://github.com/bitcrowd/frontend/issues/24

Issue #298, opened 2019-11-05
Find replacement packages for the styleguide & VRT


planktonic:
Due to issue #294, we need to replace the styleguide/component-library, and investigate our VRT options. The [SC5-Styleguide](https://github.com/SC5/sc5-styleguide) is abandoned, and the use of [Backstop-JS](https://github.com/garris/BackstopJS) should be re-evaluated following a lot of change in this space over the last few years.

What we called the ‘styleguide’ previously fulfilled the following roles:
- a component library (probably a more accurate name these days), allowing users to see the available css objects
- documentation. Each CSS object has associated explanatory comments and example code, generated from SassDoc comments in the source code
- Those examples are also rendered (not just code), so the VRTs can be run against them

Our choice of component library will likely affect our choice of VRT runner (and vice versa). One issue with using Backstop-JS is that it was a globally-installed package, not version-controlled and auto-installed with npm/yarn (Q: does it still need to be globally-installed?)

planktonic:
We’ve got a replacement styleguide ([storybook](https://storybook.js.org)), and no intention of adding a VRT system right now. If that changes, we can add a new issue for it, but for now I’ll close this.

PR #307, opened 2020-03-11
298/find replacement packages for the styleguide vrt

planktonic:
Partial fix for #298 — this adds a component library as a replacement for the previous styleguide.

- Adds storybook
- Moves the comments in `sass` files in `md` files
- Uses a storybook plugin to display the markdown next to the story
- groups the files for each CSS class into its own folder as each now has three files

Advantages:
- We’re familiar with storybook
- bitstyles is a set of UI components more than a design system, requiring a component library more than a styleguide
- Documentation is easier to read than it was in comments, both in the MD file, and in storybook
- We can run snapshot and VRT tests against the fullscreen version of each story

Disadvantages:
- More folder structure.
- Right now node-sass is falling behind the specs for sass — index imports should work (like js, where you can import from a component’s folder, and will automatically get the `index.js` file without specifying that explicitly), but they currently do not. To avoid having imports like `@import 'bitstyles/objects/list-inline/list-inline';`, this feature is required now we have per-class folders. he workaround is to name the files `_.scss` which looks odd, but will hopefully be temporary, and not so different to the equally-meaningless `index.scss`.


planktonic:
Thanks so much for your feedback @lwassermann! 🙏 

Responses, in order:
- The `base` stories/examples are there so that we can have snapshots of those elements with no classes applying styles (there’re currently still no VRT snapshots — this PR will allow that to restart)

The `base` styles themselves are very barebones, being just a few colors and text sizes that are largely inherited (and the normalize/reset css). This was an architectural decision taken when we created bitstyles, to minimise the amount of overriding we needed to do. With this approach, if you don’t apply one of our object-level input _classes_ to your input, all you get is font-sizes and text colors that match the rest of your site. This seemed a reasonable base on which to style them as you want in your current project, with minimal overriding.

It also follows the atomic/BEM/ITCSS system — applying styles only through classes in order to avoid specificity issues.

This is something I’ve wanted to revisit though — I can see it being useful having (some of?) the styles from say `o-input--text` (like padding, color overrides, border) applied as default to all input types that could be considered 'text'. Overriding those styles later on with a class would not present specificity issues (as the selector is just an element, it’s a lower specificity), and though there may be some properties that need to be re-declared, it would provide much easier out-of-the-box usage.

Any thoughts on that? Making the change would be better done in another branch anyway, I think.

- ‘The `text-align: justify` example content’ & ‘The `no-margin` stories don't show much’: Yup, that’s true. I would add both of these points to a new issue to update the examples all in one go, together (I’m sure there are other places we could improve the docs & examples).

planktonic:
I’ll merge this now, as it’s a big PR that others depend on — from feedback it seems the general approach/choice of tech is acceptable. I’ve created issues to investigate & fix the issues raised in the discussion above:

- #315 
- #316

PR #321, opened 2020-03-19
291/flex-primary-breaks-with-lots-of-content

### LINKED ISSUE #291: .l-flex__primary breaks when it contains lots of content
When using `.l-flex__primary`, if its contents are long, it overflows its parent `.l-flex`. This is because `flex-shrink` is set to zero. Better may be: `flex: 1 1 0%;` i.e. start at zero but grow to fill space, and be smaller than content when necessary. Maybe don’t need the unit, check for browser issues (iirc there are a couple of firefox flexbox bugs related to not using a unit, but that needs checking).

planktonic:
Fixes #291 

- flex__primary now only sets flex-grow — the element will no longer cause the flex parent to overflow when it’s content is of unrestrained width
- Adds the __no-shrink child class, and the --wrap modifier
- Improves the examples, and the documentation

### DIFF
--- a/scss/bitstyles/layout/flex/flex.md
+++ b/scss/bitstyles/layout/flex/flex.md
@@ -2,4 +2,12 @@
 
 ## Basic flex layout
 
-Apply to an element to simply make its children share the available width e.g. a list of links in a menu. Make one of the children `.l-flex__primary` to make it take as much space as is available i.e. all the horizontal space left after its siblings take the minimum width they need.
+**Resize your viewport for these examples**
+
+Apply `.l-flex` to an element to make its children share the available width e.g. a list of links in a menu.
+
+Give it the `.l-flex--wrap` too, if the content is a list of elements with intrinsic or fixed size e.g. a list of buttons, icons, or images. The result will mean the elements are allowed to flow onto the next line when there’s not enough space for them to be adjacent (note this will not work with elements that take all available space, e.g. paragraphs of text).
+
+Make one of the children `.l-flex__primary` to make it take all available space left over. If there is no space left over, it becomes the same size as its siblings.
+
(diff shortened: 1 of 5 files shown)


scss/bitstyles/layout/flex/flex.md
```diff
@@ -2,4 +2,12 @@
...
+**Resize your viewport for these examples**
+
+Apply `.l-flex` to an element to make its children share the available width e.g. a list of links in a menu.
+
+Give it the `.l-flex--wrap` too, if the content is a list of elements with intrinsic or fixed size e.g. a list of buttons, icons, or images. The result will mean the elements are allowed to flow onto the next line when there’s not enough space for them to be adjacent (note this will not work with elements that take all available space, e.g. paragraphs of text).
+
+Make one of the children `.l-flex__primary` to make it take all available space left over. If there is no space left over, it becomes the same size as its siblings.
+
+Apply `.l-flex__no-shrink` to a flex-child to stop it being crushed when there is not enough space for it and its siblings. This is useful with elements which have intrinsic size (e.g. images, icons), or that need to remain fully-sized (e.g. buttons).
```
lwassermann: 💯 📖

scss/bitstyles/tools/_flex.scss
```diff
@@ -13,6 +13,14 @@
   align-items: stretch;
 }
 
+@mixin flex--wrap {
+  flex-wrap: wrap;
+}
+
 @mixin flex__primary {
-  flex: 1 0 auto;
+  flex-grow: 1;
```
lwassermann: This might be a breaking change. Do we have a changelog in case we publish new versions?
Or maybe increase the version number in package.json?

planktonic: We don’t have a changelog so far — any advice how best to create one? When this batch of updates is on master, I’ll create a new release with a bumped version number, and if we have it by then, a changelog :)

PR #324, opened 2020-03-24
Chore/replace vrt

### LINKED ISSUE #298: Find replacement packages for the styleguide & VRT

planktonic:
Fixes #298 

- Adds backstopJS, and updates all the tests that existed previously

Some context/thoughts :)

Storyshots would seem to’ve been an ideal drop-in solution here, but sadly it’s very tied to react, Webpack, babel etc — and entire toolchain that we don’t use here. It would take a bunch of tools and work to get it going (I took a look into it), and given the lack of focus on non-react non-SPA systems, I do not imagine they would not break things for us in future.

BackstopJS on the other hand, does what we need it to while only introducing one dependency, is platform-agnostic, runs on CI, and has been in constant development for years. The only drawback I see is the manual nature of the test config


lwassermann:
Maybe we should switch to Circle CI here as well 🤔

planktonic:
@lwassermann thanks for taking the time to run them! Were you using `yarn test`? If so, and if you have the time, could you try running

```sh
yarn test:setup
yarn test:compare
```

Thanks! :)

I think the `yarn:test` version has no way to pass, because it checks out master to create reference images from it (that was setup to avoid checking in the reference images).

My problem right now is that running those two commands above, results in failing tests for me — some inconsistency in rendering within docker :/

planktonic:
> Maybe we should switch to Circle CI here as well 🤔

Yup that’d be great! Who do I ask? :)

lwassermann:
> ```shell
> yarn test:setup
> yarn test:compare
> ```
  report | 161 Passed
  report | 16 Failed
😞 

> I think the `yarn:test` version has no way to pass, because it checks out master to create reference images from it (that was setup to avoid checking in the reference images).
> 
🤔 

While running  the commands, I also noticed that all the generated images are now makred as changes. Maybe add the test folder to `.gitignore`?

planktonic:
@lwassermann yup, same here for the test results, thanks! So there seems to be an issue with running these in docker 🤔 I’ll look into it.

If we keep the same approach (shell script, generating reference images from master) then yes I’ll add the folder to gitignore (originally the contents of each folder were themselves gitignored, but now backstop apparently removes the folder & recreates it. Another thing to look into)

Thanks for your time! I’ll update again tomorrow, most likely

planktonic:
Closing this — as we’re now using storybook it probably makes sense to use storyshots or similar

PR #326, opened 2020-03-25
#317 Revisit normalize/reset.css

### LINKED ISSUE #317: Revisit normalize/reset.css
Investigate: do we still need normalize? Is a smaller reset acceptable in today’s browsers? If we keep normalize, it would be great to set some better defaults — with the sass variables we use to later style at the base and object levels, so we can avoid having to re-declare all these things.

planktonic:
Fixes #317 

- Updates normalize from v4 -> v8
- It’s still inlined, so that we can selectively exclude some parts e.g. properties we will be setting ourselves later


lwassermann:
👍

Issue #306, opened 2020-03-10
Revisit selects


planktonic:
`<select>` elements can now be styled quite successfully in all the modern browsers (not options yet though). It’d be great to update the styling here

planktonic:
Fixed by #327

Issue #315, opened 2020-03-18
Update examples


planktonic:
From a comment [here](https://github.com/bitcrowd/bitstyles/pull/307#pullrequestreview-373817830
):

> “The text-align: justify example content’ & ‘The no-margin stories don't show much‘”

Check through all the examples/stories to be sure we’re showing the intent, and the effect of the class in question. Adding grey backgrounds or padding to elements, or to their parent, is often useful.

planktonic:
Also look into how we can better document the classes — MSX/storybook-readme/something else. Including documenting mixins & functions, and the sass variables that can be changed per class

PR #327, opened 2020-04-01
316 update base classes

### LINKED ISSUE #316: Update base classes
Right now a lot of the base classes are relatively unstyled — just inheriting colours and fonts from html/body. All styling is added in the `.o-input--variant` classes. These styles are all specified by Sass variables so can be customised by end-users, so it would be great to shift those styles to the base level.

See discussion [here](https://github.com/bitcrowd/bitstyles/pull/307#pullrequestreview-373817830) and [here](https://github.com/bitcrowd/bitstyles/pull/307#issuecomment-600101959)

planktonic:
Fixes #316

- Turns off mergeRules in CSSNano — it 'hoists' classes, destroying the source order
- Moves input & label styling to the base-level, removes the object-level styles
- Adds styling for selects, radios, checkboxes, and fieldsets
- More opinionated but useful styes for fieldsets, inputs, and buttons. Should mean that sett
ing up your color palette is enough, no need to also override the button & input settings
- Adds some state styling that wasn’t there before (active states), and organises the setting
s files a little more clearly
- Updated base styling of links
- Updated base styling of tables

WIP:
- I’ll add some screenshots & gifs here
- Investigate if we can/should use the same color/background/border for all the input types, instead of having separate variables (and css) for all
- Is there a better way to handle the checkbox/radio with label layout? Perhaps most minimal is to give label `display: flex`, and style the children to achieve the two layouts, but what about specificity?


planktonic:
I’ll address the documentation in one PR (see #315 )

planktonic:
Also fixes #306

PR #361, opened 2021-02-16
[#352] wide range of colors in palette

### LINKED ISSUE #352: Wide range of colors in palette
Right now we provide a color palette with the base colors, and two variants of each — a `light`, and `dark`. It’d be great to provide a richer palette of colors to choose from, giving tints of each color from 0 – 100. Use the Sass function `mix()` to progressively mix each base color with 100% – 0% white (so a color at '100' is mixed with 0% white).

It may make sense to use a helper to loop over these colors and generate them, but the simpler case of providing a manually-written list would be fine too (simpler is often better).

this change may also mean we need to update how we generate the `u-color` classes — specifying a color and name, instead of just passing in color-names that exist in the palette e.g.

```scss
$foreground-colors: (
  'primary-50': palette('primary', '50');
);
```

which would also give us more control with color utility classes

planktonic:
Fixes #352

- Updates the color system, to allow much lager palettes, generating 10 tints of each color
- Shortens name of `background-color` and `color` classes to `fg` & `bg` everywhere
- Adds gray colors, mixed with a little brand-1


andreasknoepfle (approved):
Looks good from looking over everything. Can't say much about if there was something else that needs to be updated, since that is hard to see :D 

I like the shorter names that still are meaningful (fg, bg) and also the added support for larger palettes. Looks good as far as I can see.

PR #364, opened 2021-02-16
add avatar object

planktonic:
Adds an avatar object, round-clipped image.

Looks like:

Screenshot 2021-02-16 at 14 55 31


andreasknoepfle (approved):
Well done. Nice cat 💯

PR #365, opened 2021-02-16
moving settings

planktonic:
- Moves all the settings files from the centralised folder, into each of the relevant objects’ folders
- Exception is the link settings, as that needs to be available in the base-level `<a>` styles, so that needs a better solution


agatheblues (approved):
Looks good! 🚜

PR #366, opened 2021-02-16
Add badge component

planktonic:
Adds a badge component, to be used for badges, tags, filters etc.
Also contains some updates to sizing & utility classes

Looks like:

Screenshot 2021-02-16 at 17 04 12

Or with buttons for removal:

Screenshot 2021-02-16 at 17 03 59

And on hover:

Screenshot 2021-02-16 at 17 04 04


scss/bitstyles/layout/box-alignment/box-alignment.stories.mdx
```diff
@@ -0,0 +1,66 @@
...
+            elements of different height
+          </div>
+        </li>
+      </ul>
+    `}
+  </Story>
+</Canvas>
+
+
+### Center
```
agatheblues, line 40:
```suggestion
### End
```

scss/bitstyles/layout/box-alignment/box-alignment.stories.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+Set `align-items` to center to align each flex child to their vertical center:
```
agatheblues, line 42:
```suggestion
Set `align-items` to center to align each flex child to their vertical end:
```

agatheblues: Do you also need to add one for `start`?

scss/bitstyles/objects/badge/_.scss
```diff
@@ -0,0 +1,26 @@
...
+    --button-bg-hover: #{palette($variant, $bitstyles-badge-color)};
+
+    color: palette($variant, $bitstyles-badge-color);
+    background-color: palette($variant, $bitstyles-badge-background-color);
+  }
+}
+
+.o-badge__button {
+  margin-right: -#{spacing('s')};
+  margin-left: spacing('xs');
```
agatheblues, line 25: What does the minut and brackets mean here `-#{spacing('s')}` 🤔 ?

scss/bitstyles/utilities/font/_.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.#{$bitstyles-namespace}u-font--bold {
+  font-weight: $bitstyles-font-weight-bold;
+}
+
+.#{$bitstyles-namespace}u-text--italic {
+  font-style: italic;
+}
+
+.#{$bitstyles-namespace}u-text--not-italic {
```
agatheblues, line 21: Maybe `u-text--normal` is a better name here?

scss/bitstyles/utilities/font/font.stories.mdx
```diff
@@ -0,0 +1,59 @@
...
+    {`
+      <div class="u-bg--brand-2 u-padding-m">
+        <div class="u-bg--background u-font--italic">u-font--italic. Pastry marzipan gummi bears lollipop pie pie carrot cake dessert. Pie liquorice marshmallow muffin sweet roll gingerbread candy canes. Tiramisu cotton candy bonbon gingerbread brownie pastry. Croissant apple pie croissant halvah lollipop chocolate cake candy canes liquorice marshmallow.</div>
+      </div>
+    `}
+  </Story>
+  <Story name="font--not-italic">
+    {`
+      <div class="u-bg--brand-2 u-padding-m">
+        <div class="u-bg--background u-font--normal">u-font--normal. Pastry marzipan gummi bears lollipop pie pie carrot cake dessert. Pie liquorice marshmallow muffin sweet roll gingerbread candy canes. Tiramisu cotton candy bonbon gingerbread brownie pastry. Croissant apple pie croissant halvah lollipop chocolate cake candy canes liquorice marshmallow.</div>
```
agatheblues, line 55: Huh actually it appears `font--not-italic` and `u-font--normal` are mixed up over there 🍸

agatheblues (approved):
Impressive work 🏊‍♂️ ! I think I found a couple tiny issues, apart from them you can ship.

scss/bitstyles/objects/badge/_.scss
planktonic: The `-` is there because the button needs a negative margin to pull it closer to the edge of the button (so the curves match up — see the button in the screenshots). The `#{}` part is the Sass syntax for interpolation — so it’ll render `0.6rem`, in this case.

Sadly the interpolation is required in the negative version because Sass doesn’t recognise the `-spacing` function it sees in `-spacing('xs')`

scss/bitstyles/utilities/font/_.scss
planktonic: Thanks! I’ll look into this and the mixup further down 🎉

planktonic: Ah, using `normal` as the variant wasn’t possible because it’s already used for the font-weight class further up in this file — in CSS, `font-weight` can be `normal`, and so can `font-style`. In the interests of keeping the classnames shorter, it seemed better to use `italic`/`not-italic` instead of adding `-style`/`-weight` to each of them

scss/bitstyles/utilities/font/font.stories.mdx
planktonic: Amazing, thanks! This is fixed, it’s now using `u-font--not-italic` 🎉

PR #367, opened 2021-02-16
Add grouped buttons

planktonic:
- Adds grouped buttons
- margins & paddings now generated from an over-rideable list

Looks like:

Screenshot 2021-02-16 at 17 30 47
Screenshot 2021-02-16 at 17 30 59


scss/bitstyles/objects/typography/_.scss
```diff
@@ -1,11 +1 @@
-.#{$bitstyles-namespace}o-h0,
-.#{$bitstyles-namespace}o-h1,
-.#{$bitstyles-namespace}o-h2,
-.#{$bitstyles-namespace}o-h3,
-.#{$bitstyles-namespace}o-h4,
-.#{$bitstyles-namespace}o-h5,
-.#{$bitstyles-namespace}o-h6 {
-  @include generic-heading;
-}
```
agatheblues, line 9: This doesn't appear to be related to the grouped buttons, but I guess these lines have been removed for a reason?

agatheblues (approved):
Nice, love the joined buttons! 👯‍♂️

scss/bitstyles/objects/typography/_.scss
planktonic: The buttons use this `.o-h6` class to use a slightly smaller font-size from our typographic scale (the background on the buttons makes them visually larger & more clickable, so the font-size can be a little smaller).

I removed the `generic-heading` mixin here because it also adds margin-bottom, and `font-weight: bold` — it was intended to make headings in e.g. blog/article content, but actually we use the classes in UIs just for setting font-size (like the buttons here), and for overriding font-size of `<h2>/<h3>` elements to visually match designs. That meant a lot of overriding that margin and font-weight with extra classes, and even encourages use on non-semantic elements like `<div>`s. As `<h>` heading elements get the margin & font-weight anyway, it’ better that these classes are utility classes.

So the outcome of all this — I’ll move these classes to the utility layer! Thanks for bringing it up :D

PR #369, opened 2021-02-17
Bumps node version

planktonic:
Travis was complaining that the version of node set in `.travis.yml` was too low for one of the npm packages. this bumps node to v12.13.0


planktonic:
Travis is passing again, I’ll merge this

PR #379, opened 2021-02-18
Update storybook

planktonic:
Updates storybook to the latest version, and a couple of other dependencies


klappradla (approved):
Stay with the shiny things 🌟

klappradla:
How about also updating the Node version in `.nvmrc`?  It's hanging around somewhere at v12, while Node LTS is already beyond v14 😉

planktonic:
@klappradla yeh it is! Updated to the latest LTS (14.15.5)

PR #390, opened 2021-03-01
[#389] Publish static storybook to GitHub pages

### LINKED ISSUE #389: Publish static storybook to github pages

planktonic:
Fixes #389

- Adds gh-pages package, and a couple of new commands for deploying to github-pages to packagae.json
- Storybook will now default to show docs (when they exist for a component), and docs come first
- Updates storybook

Looks like: https://bitcrowd.github.io/bitstyles/

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,9 @@
     "css:stats": "analyze-css --file build/bitstyles.css --pretty > test/css-stats.json",
     "precss:graph": "yarn build",
     "css:graph": "specificity-graph build/bitstyles.css --output test/specificity_graph/",
-    "build-storybook": "build-storybook"
+    "build-storybook": "yarn build && build-storybook",
+    "predeploy-storybook": "yarn build-storybook",
+    "deploy-storybook": "gh-pages -d storybook-static"
   },
   "repository": {
     "type": "git",
@@ -43,10 +45,10 @@
   "homepage": "https://github.com/bitcrowd/bitstyles#readme",
   "devDependencies": {
     "@babel/core": "^7.8.7",
-    "@storybook/addon-actions": "^6.1.18",
-    "@storybook/addon-essentials": "^6.1.18",
-    "@storybook/addon-links": "^6.1.18",
-    "@storybook/html": "^6.1.18",
+    "@storybook/addon-actions": "^6.1.20",
+    "@storybook/addon-essentials": "^6.1.20",
+    "@storybook/addon-links": "^6.1.20",
+    "@storybook/html": "^6.1.20",
     "analyze-css": "^0.12.9",
     "autoprefixer": "^9.8.0",
     "babel-loader": "^8.1.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -17,7 +17,9 @@
     "css:stats": "analyze-css --file build/bitstyles.css --pretty > test/css-stats.json",
     "precss:graph": "yarn build",
     "css:graph": "specificity-graph build/bitstyles.css --output test/specificity_graph/",
-    "build-storybook": "build-storybook"
+    "build-storybook": "yarn build && build-storybook",
```
klappradla, line 20: Is this overriding the storybook-internal `build-storybook` command? Wonder why this does not collide 💥

package.json
```diff
@@ -17,7 +17,9 @@
...
+    "predeploy-storybook": "yarn build-storybook",
```
klappradla, line 21: How is this command used? Is it automatically triggered by some internal magic of the `gh-pages` package?

If not, I wonder if it would not make sense to integrate it with the `deploy-storybook` command to ensure it's never forgotten.

package.json
```diff
@@ -17,7 +17,9 @@
...
+    "deploy-storybook": "gh-pages -d storybook-static"
```
klappradla, line 22: Not sure about the exact workflow you had in mind for this, but one could even use a Github Action (e.g. [this](https://github.com/marketplace/actions/deploy-to-github-pages)) to automatically update these docs 🤔

klappradla (commented):
Neat 👏 

I left some minor questions here and there, but nothing keeping this from hitting 🌊.

When rolling this out, it probably also makes sense to update the README for a) how to "deploy" the docs and b) where to find them _(which could also go into the repo description)_. But both can of course also be done in a follow up PR.

package.json
planktonic: It’s magic, but from yarn/npm :D if you prepend `pre` to the name of an existing script, (`deploy-storybook` in this case, which is on the next line) yarn/npm will run that script before running the script itself when you call `yarn script-name`.

package.json
planktonic: Hmm, the name of this script was already `build-storybook`, as setup by storybook itself on install. I guess it doesn’t clash because one is run with yarn/npm, the other is a CLI tool?

package.json
planktonic: That action looks perfect, thanks! If we use it, there’d be no need for the `gh-pages` package, nor any of the deploy scripts, right? 🤔

klappradla: I didn't dive into the details of it. But I'd expect it to work like that, yes.

package.json
klappradla: 🤯 wow, thanks 🙇

package.json
klappradla: Oh, makes sense 🙈 

Thanks for explaining 🙇

klappradla (approved):
✅

planktonic:
@klappradla I updated to use that github action you linked. Is there any way to test that it works? (it’s set to run when merging to main)

PR #392, opened 2021-03-01
[#371] Add component: table layout

### LINKED ISSUE #371: Add component: table layout
We need a basic table layout with table headers  & an “actions” column where buttons can go

planktonic:
Fixes #371 

Adds a table component, with some fake content

Looks like:

Screenshot 2021-03-01 at 14 54 29

Screenshot 2021-03-01 at 14 54 21

Screenshot 2021-03-01 at 14 54 48


scss/bitstyles/ui/table.stories.mdx
```diff
@@ -0,0 +1,74 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
+
+<Meta title="UI/Data/Table" />
+
+# Table
+
+Add base styling to your tables by giving them the `o-table` class. As shown here, always add an overflow wrapper to handle smaller viewports. Tables do not resize well, so allowing horizontal scrolling is a simple and clear way to make the table accessible to smaller devices.
```
klappradla, line 7: 👏

scss/bitstyles/settings/_global.color-palette.scss
```diff
@@ -64,17 +64,17 @@ $bitstyles-color-palette: (
...
-    '80': mix($bitstyles-base-color-white, $bitstyles-base-color-background, 20%),
-    '70': mix($bitstyles-base-color-white, $bitstyles-base-color-background, 30%),
-    '60': mix($bitstyles-base-color-white, $bitstyles-base-color-background, 40%),
-    '50': mix($bitstyles-base-color-white, $bitstyles-base-color-background, 50%),
-    '40': mix($bitstyles-base-color-white, $bitstyles-base-color-background, 60%),
-    '30': mix($bitstyles-base-color-white, $bitstyles-base-color-background, 70%),
-    '20': mix($bitstyles-base-color-white, $bitstyles-base-color-background, 80%),
-    '10': mix($bitstyles-base-color-white, $bitstyles-base-color-background, 90%),
-    '5': mix($bitstyles-base-color-white, $bitstyles-base-color-background, 95%),
+    '95': mix($bitstyles-base-color-brand-1, $bitstyles-base-color-background, 3%),
```
klappradla, line 67: Just to double check: the 3% is on purpose?

scss/bitstyles.scss
```diff
@@ -54,8 +54,8 @@
 // Organisms ///////////////////////////////////
 
 @import 'bitstyles/organisms/modal/';
-@import 'bitstyles/organisms/expander';
 @import 'bitstyles/organisms/navbar/';
+@import 'bitstyles/organisms/table/';
```
klappradla, line 58: Why is this still called "organisms" while the stories where changed to "ui"?

klappradla (approved):
Awesome 💯 

While being a huge fan of rounded corners, I'm not sure about the lower part here:

Screenshot 2021-03-02 at 10 50 44

Personal perference, but I find the corners a bit irritating here.

scss/bitstyles/settings/_global.color-palette.scss
planktonic: No! Thanks, good catch 🙇‍♂️

scss/bitstyles.scss
planktonic: There’s a plot between the “UI” documentation and the components beneath — the UI-level docs are mostly examples of pre-built components that often compose a bunch of atoms, organisms, and utilities. They don’t necessarily nor often correspond to a separate organism.

Tables and navbars need their own CSS to work (and some utilities etc), so they exist as organisms in the CSS, and like all the CSS classes they’re documented in storybook, in a manner which shows the classes and child classes, but not necessarily how best to use them. The documentation on tables at the UI level shows how to use the tables in a roughly real-world instance, but doesn’t show you examples of each of the child elements, classes etc.

That’s because each of those layers is aimed at a different person and/or use case. Does that make sense, any downsides or problems that you see? I’m very sure that I’ll need to do a cleanup round to be sure that structure is implemented well across all examples & classes, as my ideas on this have developed quite a lot as I’ve been working on the changes

scss/bitstyles/settings/_global.color-palette.scss
planktonic: Aha, I checked this — yes it’s deliberate, though it certainly looks like an error. I’m not sure how to handle it better right now 🤔 I’ll add a comment.

TLDR: Because nothing related to color and color perception is easy, if I mix 5% of the brand color with white, the output is still very dark — switching to 3% makes a large difference, and is much closer to the background color, as would be expected I think

planktonic:
@klappradla I tweaked the padding, so that buttons on the right fit the bottom-right corners a little better. Looks like this:

scss/bitstyles.scss
klappradla: Makes sense, thanks for explaining 🙇

klappradla:
For the table padding: the sceenshot isn't visible for me, but it looks good on the deployed docs ✔️

PR #409, opened 2021-03-09
[#407] Add component: header bars

### LINKED ISSUE #407: Add component: header bars
A heading element on the left, with action button(s) on the right, collapsing into a single column on smaller screens.

One is already used in the page header component, but doesn’t exist on its own. A useful component for sub-headings too, so break it out into new molecule.

planktonic:
Fixes #407 

Adds some section headers, with variants labels, badges, and buttons.

Looks like:

Screenshot 2021-03-09 at 17 36 08


nickgnd (approved):
Nice 👏 
Really well done! 💯

PR #411, opened 2021-03-10
[#410] Equalise button borders

### LINKED ISSUE #410: Equalise button borders
So that they fit next to each other (without being stretched by flex or something), all the buttons should take their button width from the base button class, changing only the border-color in each variant

planktonic:
Fixes #410 

All buttons now only set their border-color to override the base button styles, and for each of their hover/active/etc states. This simplifies the buttons a little, avoids the case where border-width is changed on hover/etc, and makes sure that all button types can be placed next to each other and have equal heights.

Not really any visual difference (the border is a _tiny_ bit thinner, as it’s based on font-size now), but here’s some of the buttons:

Screenshot 2021-03-10 at 10 19 19

Screenshot 2021-03-10 at 10 19 26

Screenshot 2021-03-10 at 10 19 59

Screenshot 2021-03-10 at 10 19 11

### DIFF
--- a/scss/bitstyles/atoms/button/settings.scss
+++ b/scss/bitstyles/atoms/button/settings.scss
@@ -24,7 +24,7 @@ $bitstyles-button-box-shadow: 0 0 spacing('xs') rgba(palette('brand-1', '80'), 0
 
 $bitstyles-button-color-hover: palette('white') !default;
 $bitstyles-button-background-color-hover: palette('brand-1') !default;
-$bitstyles-button-border-hover: 2px solid palette('brand-1') !default;
+$bitstyles-button-border-color-hover: palette('brand-1') !default;
 $bitstyles-button-box-shadow-hover: 0 0 spacing('s') rgba(palette('brand-1'), 0.2) !default;
 
 //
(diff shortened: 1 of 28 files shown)


scss/bitstyles/atoms/button/settings.scss
```diff
@@ -16,29 +16,30 @@ $bitstyles-button-icon-spacing: spacing('s') !default;
 
 $bitstyles-button-color: palette('white') !default;
 $bitstyles-button-background-color: palette('brand-1', '80') !default;
-$bitstyles-button-border: 2px solid palette('brand-1', '80') !default;
+$bitstyles-button-border-width: spacing('xxxs') !default;
```
tessi, line 19: 🙈

tessi (approved):
looks good 🧑‍🎨

scss/bitstyles/atoms/button/settings.scss
planktonic: soon: `xxxxxxxxxxxxxxxxxxxxxxxxs`

PR #413, opened 2021-03-10
[#406] Add component: notifications

### LINKED ISSUE #406: Add component: notifications
Notification cards to appear top-right, to inform the user of an action that has happened (e.g. “Password updated successfully”), with an icon on the left.

Variants:
- dismiss button top right
- ok/cancel buttons on the bottom
- various icon colors

planktonic:
Fixes #406 

Adds a simple notification component, and a notification with buttons.

Looks like:

Screenshot 2021-03-10 at 16 34 11

Screenshot 2021-03-10 at 16 34 20

### DIFF
--- a/scss/bitstyles.scss
+++ b/scss/bitstyles.scss
@@ -43,6 +43,7 @@
 @import 'bitstyles/atoms/card/';
 @import 'bitstyles/atoms/content/';
 @import 'bitstyles/atoms/topbar/';
+@import 'bitstyles/atoms/flash/';
 
 //
 // Organisms ///////////////////////////////////
--- a/scss/bitstyles.scss
+++ b/scss/bitstyles.scss
@@ -51,6 +51,7 @@
 @import 'bitstyles/organisms/navbar/';
 @import 'bitstyles/organisms/table/';
 @import 'bitstyles/organisms/ui-group/';
+@import 'bitstyles/organisms/notification-center/';
 
 //
 // Utilities ////////////////////////////////////
--- a/scss/bitstyles/atoms/button--ui/_.scss
+++ b/scss/bitstyles/atoms/button--ui/_.scss
@@ -12,7 +12,6 @@
   &:hover,
   &:focus {
     color: $bitstyles-button-ui-color-hover;
-    text-decoration: none;
     background: $bitstyles-button-ui-background-color-hover;
     border-color: $bitstyles-button-ui-border-color-hover;
     box-shadow: none;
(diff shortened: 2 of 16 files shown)


scss/bitstyles/atoms/button--icon-reversed/_.scss
```diff
@@ -14,23 +14,23 @@
...
   }
 
   &:disabled {
     color: $bitstyles-button-icon-reversed-color-disabled;
     cursor: default;
     background: $bitstyles-button-icon-reversed-background-color-disabled;
-    border: $bitstyles-button-icon-reversed-border-disabled;
+    border-color: $bitstyles-button-icon-reversed-border-color-disabled;
     box-shadow: $bitstyles-button-icon-reversed-box-shadow-disabled;
   }
```
andreasknoepfle, line 35: So does that mean that we are now adding a border to all the buttons--icons and button--dangers that was not there before?

andreasknoepfle (approved):
The notification itself looks great :) 
This is meant to be displayed in the top-right corner, right? 

While this is great for those SPA-style apps, to display events, there are still some use-cases for classic web pages, where we might display the notification more present, since it is the result of an immediate action after a page load. 

Examples: 

- Login fails --> We do not want to show exactly what failed, for security reasons 
- Buttons without form that can still fail (certain actions make it impossible to hide the button if it would result in an error, beforehand). I know one wants to avoid this, but in real life admin backends this is normal (network issues, conflicts, other....)

scss/bitstyles/atoms/button--icon-reversed/_.scss
planktonic: Aha, no that’s from #411 — that got rebased but I didn’t update this branch. I’ve rebased! That PR will need to be merged before this one

(The change here is that border-color can be changed on hover etc, but all buttons will have the same _width_ of border. This means sizing and vertical alignment is equal, and no shifting happens on hover/etc)

planktonic:
@andreasknoepfle yes this a top-right kinda notification. It’s meant for any events, but if the event is form-related or occurs after a page reload, a “flash”-style would be more noticeable 👍 

I’ll add a flash component in this branch, and a couple of examples:

[] - global flash, for “Login failed” etc
[] - flash in a form, to indicate something was invalid (field-specific feedback like “Email address invalid” should still render next to the relevant field though)

I’ll assume for now that there will be no “dismiss notification” button. Did I miss anything from that list?

planktonic:
I’ve added a flash component, looks like:

Screenshot 2021-03-15 at 13 00 02

Screenshot 2021-03-15 at 13 00 08

andreasknoepfle (approved):
Absolute awesomeness 💯

PR #417, opened 2021-03-11
[#416] Fix webfont font-weights

### LINKED ISSUE #416: Fix webfont font-weights
Possibly they were caching. The webfont is only available at regular weight

planktonic:
Fixes #416 (fonts were possibly caching? but failed to load)

- Fixes the specifications of webfont weights
- renames the font files to match convention
- Improves the documentation

Looks like:

Screenshot 2021-03-11 at 12 46 43


scss/bitstyles/base/font-face/_.scss
```diff
@@ -25,15 +25,15 @@
...
+      #{$bitstyles-webfont-font-url + '-' + $bitstyles-font-weight-medium}
     );
   }
   @if (list.index($bitstyles-webfont-variants, 'bold')) {
     @include webfont(
       $bitstyles-font-family-webfont,
       normal,
       $bitstyles-font-weight-bold,
-      $bitstyles-webfont-font-url
+      #{$bitstyles-webfont-font-url + '-' + $bitstyles-font-weight-bold}
```
andreasknoepfle, line 36: Wouldn't it make more sense if the url would all be variables and you would construct them in `typography`

from a `$bitstyles-webfont-font-url-base`? to set the defaults? Like this one could override this with fonts that do not match the naming patterns if one wishes :)

andreasknoepfle (approved):
Quick suggestion. Otherwise lgtm

scss/bitstyles/base/font-face/_.scss
planktonic: Yeh this part was bugging me too. Thanks for the poke, I’ve updated it — I’m now specifying a set of fonts with their properties as a sass map, and this file iterates over it. If the map is empty, no `@fontface`s are output

PR #424, opened 2021-03-16
[#422] SVG icons

### LINKED ISSUE #422: SVG icons
Use `<use>`

planktonic:
Fixes #422 

- Adds an icons.svg “sprite” file which holds all the icons in a centralised location for easier maintenance
- Replaces all icons with SVG `<use>`
- Tweaks & equalizes the stroke width of icons

Looks almost same as it did before, this is mainly just code changes:

Screenshot 2021-03-16 at 12 28 24


tessi (approved):
love this change, we've done similar in client apps already 🍏 
it's especially great because it gives icons a name instead of just that (to humans) unreadable path.

PR #426, opened 2021-03-17
[#425] More permissive button states

### LINKED ISSUE #425: More permissive button states
After a little usage in a project:

Right now the button aria-related states are too restrictive. `.a-button--tab` will need to respond to `aria-current`, not just `aria-selected`. `.a-button--nav`  and `nav--large` will need to respond to `aria-expanded`

Dropdown styling should include the border by default, as the element is used by simply `display:none` or not.

planktonic:
Fixes #425 

A bunch of extra states for buttons, so that we can use certain button types in different situations, which fixes e.g. highlighting current page in pagination:

- `.a-button--nav` and `.a-button--nav-large` both now respond to `aria-expanded="true"`
- `.a-button--tab` and `.a-button--ui` now respond to `aria-current`
- `.a-button--ui` border is now more subtle
- `.a-button--danger` no applies no colors until hover, so it can be used with any button class

Also fixes a couple of issues found while implementing these components in a live project:

- `.a-dropdown` border is now always applied, as we show/hide by adding `display: none` with JS
- `.a-navbar` is now rendered over other content (z-index)

Looks like:

Screenshot 2021-03-17 at 13 51 25

### DIFF
--- a/scss/bitstyles/ui/joined-buttons.stories.mdx
+++ b/scss/bitstyles/ui/joined-buttons.stories.mdx
@@ -61,6 +61,8 @@ For a series of buttons that are very directly related e.g.
   </Story>
 </Canvas>
 
+Mark the current page in a series of pagination buttons, by giving the link the `aria-current="page"` attribute.
+
 <Canvas>
   <Story name="Joined buttons - pagination">
     {`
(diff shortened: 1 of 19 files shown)


scss/bitstyles/ui/joined-buttons.stories.mdx
```diff
@@ -79,7 +81,7 @@ For a series of buttons that are very directly related e.g.
           </a>
         </li>
         <li class="u-flex">
-          <a href="/" class="a-button a-button--ui o-ui-group__item u-round-0 u-h6">
+          <a href="/" class="a-button a-button--ui o-ui-group__item u-round-0 u-h6" aria-current="page">
```
tessi, line 84: 😍

tessi (approved):
looks great. can't wait to see these additions in our current project

Issue #436, opened 2021-03-29
Add npmignore file


planktonic:
Add a `.npmignore` file with the same contents as the `.gitignore`, but without `build`. This way, the built resources can be made available via CDN etc later

PR #439, opened 2021-03-29
[#436] Add npmignore file (2)

planktonic:
Removes the npmignore, in favour of the build key in package.json


andreasknoepfle (approved):
Looks better, hope it is bundled now :)

PR #442, opened 2021-04-01
[#423] Improve storybook docs

### LINKED ISSUE #423: Improve storybook docs
- [x] The components in the UI layer are most often built using a collection of atoms, molecules, organisms. List these “dependencies” in the documentation as links, so readers can drill down to those.
- [x] Grids in the intro page do not have gap classes applied
- [x] make sure the navbar (UI layer) mentions clearly how to mark the active page
- [x] add example(s) of selects next to inputs, buttons etc, so we can check the heights match
- [x] utility classes should be listed in storybook by the css property they specify

planktonic:
Fixes #423 

- Adds gap classes where they were missing
- Started adding 'also uses' sections linking to the classes used in the UI layer examples
- Upgrade storybook, adds `@storybook/addon-postcss` to appease the warnings on running storybook (implicit postcss plugin will not be supported at next major version).


planktonic:
- Lists all CSS dependencies of a component at the UI layer
- Fixes intro layout
- navbar mentions active page
- shows buttons, selects, and inputs adjacently
- utility classes listed by css property in storybook

Looks like:

Screenshot 2021-04-06 at 15 11 54

Screenshot 2021-04-06 at 15 12 10

klappradla:
I cannot `build-storybook` (and therefore not look at the changes) on this branch:
```
ERR! Module not found: Error: Can't resolve '@storybook/api/dist/lib/merge'
```

planktonic:
@klappradla thanks for taking a look! For some reason (related to updating to storybook, I think) you need to delete `node_modules` then `yarn install`, then run storybook

tessi (approved):
wow, it's a lot. but looks good

PR #444, opened 2021-04-06
[#443] Horizontal navbar with many links

### LINKED ISSUE #443: Horizontal navbar with many links
- Add a story for the navbar with many links
- Improve the way it handles those, so the wrapping makes sense (perhaps scrolling)

planktonic:
Fixes #443 

- Tweaks the navbar UI component to allow the list of links to grow safely, by overflowing.
- Navbar width is no longer limited to content-width

Looks like:

Screenshot 2021-04-06 at 15 54 21

Screenshot 2021-04-06 at 15 54 48

Screenshot 2021-04-06 at 15 57 58


scss/bitstyles/organisms/navbar/_.scss
```diff
@@ -1,4 +1,4 @@
-.c-navbar {
+.o-navbar {
```
andreasknoepfle, line 1: I guess that means we have to reflect the changes in the projects, right? 

Would it make sense to start writing CHANGELOG entries for things like this (breaking changes at least)

scss/bitstyles/ui/navbar.stories.mdx
```diff
@@ -95,3 +95,94 @@ Give the active link the `aria-current="page"` attribute.
     `}
   </Story>
 </Canvas>
+
+## Handling large numbers of links
+
+This navbar is only intended to hold a few link to the main sections of your site. If you absolutely must have more links than can be easily handled, the navbar will scroll:
```
andreasknoepfle, line 101: 👍

andreasknoepfle (approved):

scss/bitstyles/organisms/navbar/_.scss
planktonic: Yes totally 👍 Any classname change or removal is a breaking change (though I don’t think we should  bump the major version number). I’ll start using the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format for the commit messages on merge to main, hopefully that’ll help make sure we don’t miss any when compiling a release. Should I also add a changelog.md file?

scss/bitstyles/ui/navbar.stories.mdx
planktonic, line 101:
```suggestion
This navbar is only intended to hold a few links to the main sections of your site. If you absolutely must have more links than can be easily handled, the navbar will scroll:
```

scss/bitstyles/organisms/navbar/_.scss
andreasknoepfle: I think this could be quite nice :) 
M_alte did add one to chromic_pdf and the format is quite good. 
https://github.com/bitcrowd/chromic_pdf/blob/main/CHANGELOG.md

andreasknoepfle: Maybe even have a `Breaking` section for breaking changes.

PR #445, opened 2021-04-06
[#418] Add component: sidebar

### LINKED ISSUE #418: Add component: sidebar
Navbar variant that’s side-aligned

planktonic:
Fixes #418 

Basic layout for a sidebar
    
~~TODO~~ Done:
- [x] What happens on smaller screens?
- [x] Add an example with lots of links, and add scrolling
- [x] Add an example with long link text, and handle width/clipping/etc

Looks like:

Screenshot 2021-04-07 at 12 00 52

Screenshot 2021-04-07 at 12 11 21


scss/bitstyles.scss
```diff
@@ -77,6 +78,7 @@
 @import 'bitstyles/utilities/items/';
 @import 'bitstyles/utilities/self/';
 @import 'bitstyles/utilities/justify/';
+@import 'bitstyles/utilities/height/';
```
andreasknoepfle, line 81: would it make sense to eventually sort these imports by name :D After all it is a lot by now

andreasknoepfle (approved):
Awesome stuff 👍  Can't wait to see it in action :)

scss/bitstyles.scss
planktonic: I’ve been thinking about that. It’ll take a little investigation — some of these need to come before others to allow certain conventions to hold true (`.u-hidden` can be overridden at specified screensizes using e.g. `.u-block@m` or `.u-flex@l`, but that only works if `hidden` comes first in source order).

I’ll add a ticket for it :)

andreasknoepfle: If some need to come before than those can just be sorted at the top and have a space in between, to indicate that they need to come before. But does not need to happen in this ticket :)

planktonic: https://github.com/bitcrowd/bitstyles/pull/452

PR #452, opened 2021-04-08
[#446] Alphabetize imports?

### LINKED ISSUE #446: alphabetize imports?
Right now each of the atoms/organisms/utilities is imported in either random order (within their layer), or in a specific order intended to allow overriding of earlier styles. Can we alphabetise some or all of it, so it’s more ordered? How can we handle the instances where order is important?

planktonic:
Fixes #446 

- Orders the imports in the various `all.scss` files alphabetically
- alphabetize all imports in the main manifest file
- add a new `display` collection of classes, to encapsulate the order-dependency of those classes in one file, and allow the alphabetization
- Updates the docs & examples for the new display classes

Mainly an invisible change, but the new docs look like:

Screenshot 2021-04-08 at 16 36 52


scss/bitstyles.scss
```diff
@@ -21,70 +21,69 @@
 //
 // Atoms //////////////////////////////////////
 
-@import 'bitstyles/atoms/list-reset/';
-@import 'bitstyles/atoms/icon/';
+@import 'bitstyles/atoms/avatar/';
```
klappradla, line 24: Hm, it probably makes sense to add a rule for `stylelint` or `eslint` to make sure this is also adhered to after this PR is merged 😉

klappradla (approved):
Looking good in storybook 💅 

I'd suggest to add a linting rule for the imports, so that this manual effort also persists.

scss/bitstyles.scss
planktonic: Ooo, great idea 💡✨I’ll add a ticket to look into that

planktonic: #460

PR #457, opened 2021-04-13
[#455] Add fonts and icons to the release?

### LINKED ISSUE #455: Add fonts and icons to the release?
We need access to the fonts and icons so we can use them e.g. in the bitstyles_phoenix docs. If we add them to the build and therefore include them in the release, they would then be available on jsdelivr. Can we achieve that without including fonts in the release though? (to save the downloads and package size)

planktonic:
Fixes #455

- Moves icons & webfonts to a root assets folder
- Adds root assets folder to release
- Updates all the references to fonts and icons


lwassermann (approved):
Does the font license require us to credit the author somewhere? Or something like that? And even if not, maybe add a link from the README?

👍

andreasknoepfle (approved):
Can we still have this in the 1.1.0 release :) Would be awesome 🥳 

I also think an attribution for the font would not hurt.

planktonic:
Yup good point! I added a font attribution 👍

PR #481, opened 2021-05-17
[#454] Rename fixed-ratio to aspect-ratio

### LINKED ISSUE #454: Rename `fixed-ratio` to `aspect-ratio`
The new CSS property is named `aspect-ratio`, we should follow that. Also update to use the new property in supporting browsers, if that makes sense already

planktonic:
Fixes #454 

- Renames the `fixed-ratio` class and associated Sass tools to `aspect-ratio`
- Updates the mixin to use the new CSS property `aspect-ratio`, with a fallback for non-supporting browsers
- Renames the Sass variable that sets this, so naming of each required aspect-ratio is possible

Delete if not applicable:

Looks like (same as before):

Screenshot 2021-05-17 at 13 09 48


scss/bitstyles/tools/_aspect-ratio.scss
```diff
@@ -0,0 +1,42 @@
...
+// **Enforces an aspect ratio on this element and its contents**.
+//
+// An element that has intrinsic shape — width will be 100%, height will be limited
+// to keep your aspect ratio. Pass in width & height parameters to set your aspect
+// ratio.
+//
+// e.g.
+//
+// ```scss
+// .u-aspect-ratio--16-10 {
```
ammancilla, line 12:
```suggestion
// .u-aspect-ratio-16-10 {
```
The [changelog says](https://github.com/bitcrowd/bitstyles/pull/481/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR14) `and the double-dashes are replaced with single`, shouldn't this example use single dashes?

ammancilla (commented):
I am not familiar yet with how the release/versioning process of bitstyles works but since there are breaking changes it seems to be a major version bump missing. Will the version bump happen in a later stage or we don't do versioning for bitstyles?

scss/bitstyles/tools/_aspect-ratio.scss
planktonic: Yes! Thank you 🙏

planktonic:
@ammancilla thanks for the review! 💚 There are two answers to your question:

Yup, semantic versioning doesn't seem to fit CSS libraries so well, unless you accept that v100 will be reached pretty soon :D (“breaking changes” for CSS means any change to a classname, as the classnames provided are essentially the “API” that’s exposed for developers using it). So instead of incrementing major version on every breaking change, we increment minor instead. So far that seems ok, but it’s definitely open to feedback & change :)

And, updating version number in package.json etc doesn’t happen on a PR-level, but when we make releases. A release is like a snapshot of the main branch, but the main branch itself doesn’t have a version number. Just before release, we’ll create a PR that increments the version number based on the changes that have been merged to main since the last release, and updates the changelog. We do it this way around so that we know which changes are contained in a release (there could be instances where a PR hangs around for a while before being merged, and a release is made without it, so a version number in the PR wouldn’t be correct any more). Does that make sense?

ammancilla (approved):
Thanks @planktonic for the detailed answers to my questions, everything makes sense from my point of view.  Code changes looks good too so --> ✅.

ammancilla:
> So instead of incrementing major version on every breaking change, we increment minor instead. So far that seems ok, but it’s definitely open to feedback & change :)

Forgot to ask: When do we increment major versions then? 😃

planktonic:
Yeh we didn’t get that far yet 😆 My first thought would be along the lines of “we’ll know it when we see it”, but we should probably come up with a clearer definition than that. Suggestions welcome!

PR #482, opened 2021-05-18
[#474] Generate margins & padding from a sass variable

### LINKED ISSUE #474: Generate margins & padding from a sass variable
Right now the margins & paddings are generated somewhat manually. It would be great to be able to define what gets generated using a sass map (similar to the typographic scale). Iterate over breakpoints and sizes (including negative) at all four directions and convenience `x` and `y`.

If possible, DRY up the output of margins/paddings with no media query, and those with MQs (so there’s not two repeated blocks to achieve that), and maybe even output margins & paddings using the same Sass (just change margin/padding)

planktonic:
Fixes #474 

- converts the margin & padding utility classes to a variable-specified, generated format
- classnames for margins & padding utility classes no longer use the `--` notation
- stories are now comprehensive for the available padding/margin classes

✨ Will  be used by several other sets of classes, to allow more customisation and consistency ✨

Delete if not applicable:


CHANGELOG.md
```diff
@@ -8,6 +8,12 @@
 
 - Dropdown atom has a new variant `.a-dropdown--full-width` that causes the dropdown to span the full width of its nearest `position: relative` parent element. Combinable with `a-dropdown--top`
 - Gap sizes can now be specified, and the class suffixes named, using a Sass variable
+- You can now override which directions the margin and padding utility classes are output for, and what those classes are named independently of the CSS property, using `$bitstyles-margin-directions` and `$bitstyles-padding-directions` respectively
```
klappradla, line 11:
```suggestion
- You can now override which directions the margin and padding utility classes are output for (`top`, `bottom`, etc.), and what those classes are named independently of the CSS property, using `$bitstyles-margin-directions` and `$bitstyles-padding-directions` respectively
```

"Direction" wasn't obvious to me on the first read.

scss/bitstyles/utilities/margin/margin.stories.mdx
```diff
@@ -4,46 +4,220 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
...
+  'top': ('top'),
+  'right': ('right'),
+  'bottom': ('bottom'),
+  'left': ('left'),
+  'x': ('right', 'left'),
+  'y': ('top', 'bottom')
+);
+```
+
+The `'': null` entry is needed to render the margin classes with no directional modifiers.
```
klappradla, line 152: Could this also be done with a default value for the function parameter? Plus: why is there no example for this?

scss/bitstyles/utilities/padding/padding.stories.mdx
```diff
@@ -4,48 +4,165 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
...
+  'top': ('top'),
+  'right': ('right'),
+  'bottom': ('bottom'),
+  'left': ('left'),
+  'x': ('right', 'left'),
+  'y': ('top', 'bottom')
+);
+```
+
+The `'': null` entry is needed to render the padding classes with no directional modifiers.
```
klappradla, line 97: Same here: could this be done via a default value plus why is there no example?

klappradla (approved):
Impressive and great cleanups 💯 

I personally only see the benefits of a) consistency and b) using functions to clean up duplication. The part about "overriding" the directions is not obvious to me yet 🙈. Why would I for instance change `top` to point to "right" instead of "top"? Maybe I'm missing context here though...

scss/bitstyles/utilities/margin/margin.stories.mdx
planktonic: The examples for these (this and the padding one you flagged too) are all the examples that came before — the classes above all change the size of the margin, but are not directional (they apply to all directions). Since that’s not clear, and not having it _is_ inconsistent, I’ll add it as the first example in the following block of examples. Same for the padding. Thanks!

planktonic: Ok I added the all-directional examples here too ✨

Your other question about doing this with a default parameter: I wanted this to all be as configurable as possible, so that only classes you need are output (it’s common to avoid using margin-top, for example). It would be inconsistent to not allow the all-directional versions to be configured. But it did result in this slightly odd config 🤔 I’ll think on it, maybe your suggestion will work for all cases we need it to, or maybe there’s just a better way to do this

scss/bitstyles/utilities/padding/padding.stories.mdx
planktonic: (See thread over on https://github.com/bitcrowd/bitstyles/pull/482#discussion_r642394268)

scss/bitstyles/utilities/margin/margin.stories.mdx
planktonic: I’m going to go with this approach for now, and reconsider when I know how we end up using these functions and configurations — many of the utility classes could be better-built using generators. Thanks for your feedback, it was super helpful!

PR #513, opened 2021-06-29
Update dependencies

klappradla:
This fixes [CVE-2021-32723](https://github.com/advisories/GHSA-gj77-59wh-66hg)

**Changes**

- Upgrade storybook packages
- Force resolution of prismjs

We can also leave this PR open until the update can be done without `resolution` 🤔

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9,17 +9,17 @@
   dependencies:
     "@babel/highlight" "^7.10.4"
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4", "@babel/code-frame@^7.12.13", "@babel/code-frame@^7.5.5", "@babel/code-frame@^7.8.3":
-  version "7.12.13"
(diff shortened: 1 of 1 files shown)


planktonic (approved):
Thanks for this! Everything I tested, worked. But I guess it’s a little risky forcing the update. As we only publish Sass and CSS, and none of the JS dependencies reaches the end user, it may not be necessary? 🤔 I’m also unsure though, and I’m happy for this to be merged :D

klappradla:
Makes sense 👍 

How about I remove the second commit (which forces a resolution) and we for now only merge the small updates to storybook dependencies? We can then wait for an easier fix for the `prismjs` problem.

planktonic:
@klappradla yup sounds good 👍

klappradla:
Done @planktonic ✔️

planktonic:
@klappradla thank you! 💚

Issue #53, opened 2016-04-07
Unit testing for sass functions


planktonic:
e.g.
- https://github.com/ryanbahniuk/sassaby
- https://github.com/oddbird/true

PR #518, opened 2021-06-30
[#505] Allow override of gray palette mix

### LINKED ISSUE #505: allow override of gray palette mix
Right now the `gray` colors are mixed to fixed values from black and the brand-1 color. Allow that to be set using a sass variable

planktonic:
Fixes #505 

- Quantity of the highlight color mixed into the grays, and the highlight color itself, is now set by variable
- adds `shades` palettes — mixing black into the base color — to compliment most of the existing palettes (which are `tints`, mixing white into the base color)
- Adds a generator for building color-palettes, uses it for all the simpler colors in the palette
- Switches to using Sass `@use` for the color settings

👀 see `gray` palette, third row down:

|Default colors|Overridden gray|
|--|--|
|default gray|overridden gray|

(⚠️ I’m building this palette storybook page in another PR, the screenshot is here so there’s something to see!)

Delete if not applicable:


scss/bitstyles/tools/_color.scss
```diff
@@ -0,0 +1,52 @@
...
+//
+// @param {String} $mix-color  The color to mix into the base color at different percentages
+//
+// @param {list} $percentages  A list of decmial values representing the percentages at which the mix-color will be mixed into the base-color
+
+@function generate-palette($base-color, $mix-color, $percentages) {
+  $color-map: ();
+
+  @each $percentage in $percentages {
+    $color-map: map.merge($color-map, ('#{$percentage}': mix($mix-color, $base-color, 100% - percentage(math.div($percentage, 100)))));
```
tessi, line 18: ✅  reads good. I wonder if there is such a thing as tests for the logic-parts of sass, but I guess that is not a thing 🤔

scss/bitstyles/tools/_color.scss
```diff
@@ -0,0 +1,52 @@
...
+// **Returns requested colours from the global `$palette` map.**
+// e.g. `color: palette('text');`
+//
+// @param {String} $color  Base colour within the palette
+//
+// @param {String} $percent   Optional. Percentage of the color to mix with white. Defaults to `100%` (fuly the specified color, no white)
+//
+
+// stylelint-disable at-rule-no-unknown
+@function palette($color, $percent: '100') {
```
tessi, line 35: hmmm, why do we actually pre-compute percentage values and can't do that math here instead? This would allow us to use odd percentage values (98.5%) and not rely on the preconfigured percentage steps.

tessi (approved):
its a wonder to me what you can do in sass, the code reads good - just wondering about the percentage steps. but nothing blocking this PR

scss/bitstyles/tools/_color.scss
planktonic: Having the preset percentages is a feature :) We’re providing a limited palette of colors, so that we don’t e.g. end up with 20 different colors for borders because they’re set manually every time. If you need to override the percentages, there’s the `$bitstyles-color-mix-percentages` variable, and if you need to specify exact colors instead of generated colors (because of branding for example), then the map can be overridden manually.

None of which means it wouldn’t be valid to switch to a model of generating colors on-demand instead of in advance, but it probably does mean there isn’t really an advantage 🤔 wdyt?

scss/bitstyles/tools/_color.scss
planktonic: Our oldest open issue is related to sass unit tests :D (#53) It felt like overkill at the time, but now that we’re generating more CSS using functions, it could be worth investigating again 👍

Issue #396, opened 2021-03-03
Add design tokens


planktonic:
Add a section/page (as appropriate) to storybook for:
- colors
- typography
- spacing

PR #519, opened 2021-07-01
[#396] Add design tokens

### LINKED ISSUE #385: Storybook should build & watch the css too
When running `yarn storybook`, it’s currently required to also run `yarn watch` (or at least `yarn build`, once) in another terminal. It’d be great to get storybook to watch the sass files too, and use the same build.

planktonic:
Fixes part of #396 (colors)
Fixes #385 

- Storybook now builds the Sass itself with webpack, so it can get access to the sass variables
- outputs the entire color-palette sass variable to icss so its available in JS
- renders the palettes in storybook

Looks like:

Firefox_Screenshot_2021-07-05T12-04-16 222Z

Delete if not applicable:


.storybook/main.js
```diff
@@ -14,5 +16,30 @@ module.exports = {
...
+        loader: "css-loader",
+        options: {
+          importLoaders: 1,
+          modules: {
+            compileType: "icss"
+          }
+        }
+      },
+      {
+        loader: "sass-loader"
```
klappradla, line 37: Not entirely sure I understand this 🙈 

Does this enforce a specific processing order? Otherwise, why can't the `Sass` loader be part of the `css-loader` (see [css-loader docs](https://webpack.js.org/loaders/css-loader/#importloaders))?

scss/bitstyles/ui/Colors.stories.js
```diff
@@ -0,0 +1,76 @@
...
+
+  Object.keys(colorPalette).map(key => {
+    if (counter !== 0 && counter % 18 === 0) {
+      final.push(portion);
+      portion = {};
+    }
+    portion[key] = values[counter];
+    counter += 1;
+  });
+  final.push(portion);
```
klappradla, line 47: 🤯

klappradla (approved):
Looking good on storybook 💅

scss/bitstyles/ui/Colors.stories.js
planktonic: Yeh, I suspect I would react the same in a couple of months! I’ll make it a function, and add comments to hopefully make this clearer 👍

.storybook/main.js
planktonic: I’m no webpack expert, but this is what I know as our standard setup 🤔 Yes it’s ordered — sass gets compiled to css, then handed to the style-loader to insert into the page. I never knew css-loader could itself use loaders, I’ll look into it! What’s the advantage?

planktonic: Looking into this, there is a lot of confusion over what that option does! It’s not for compiling sass though — that should be done with sass-loader directly, as we’ve got here already. See https://github.com/webpack-contrib/css-loader/issues/228#issuecomment-312885975 and https://github.com/webpack-contrib/css-loader/issues/765 for some explanation, but the TL;DR is that it’s to help CSS-loader do the right thing when it encounters unresolved `@import` statements

PR #521, opened 2021-07-01
[#514] Make a-card specifiable with variables

### LINKED ISSUE #514: Make a-card specifiable with variables
add the ability to apply the atom at breakpoints

planktonic:
Fixes #514 

- Adds a generator for classes that output blocks of CSS (instead of single-property classes)
- Uses the generator to output the `.a-card` classes at breakpoints specified by sass variable

Looks like:

Screenshot 2021-07-01 at 17 27 48

Delete if not applicable:


scss/bitstyles/atoms/card/settings.scss
```diff
@@ -4,3 +4,4 @@ $bitstyles-card-box-shadow:
   0 spacing('xs') spacing('m') rgba(palette('brand-1', '100'), 0.1),
   0 spacing('xxxs') spacing('xxs') rgba(palette('gray', '90'), 0.04),
   0 0 spacing('xxs') rgba(palette('gray', '90'), 0.1) !default;
+$bitstyles-card-breakpoints: ('base', 'm', 'l') !default;
```
andreasknoepfle, line 7: So if I get this right, `base` always needs to be there otherwise we have no `a-card` at all, rigth? Would it maybe make  sense to use `list.join` to add `base` (not sure if this is valid Sass 🙈 )
```
@each $breakpoint-alias in list.join('base', $bitstyles-card-breakpoints) {
  @if $breakpoint-alias == 'base' {
    @include output-block(
      $object-level: 'a',
      $classname-root: 'card'
    ) {
      @include card;
    }
  }

  @else {
    @include media-query($breakpoint-alias) {
      @include output-block(
        $object-level: 'a',
        $classname-root: 'card',
        $breakpoint-suffix: '\\\@#{$breakpoint-alias}'
      ) {
        @include card;
      }
    }
  }
}
```

scss/bitstyles/tools/_directional-properties.scss
```diff
@@ -122,3 +122,21 @@
...
+      $bitstyles-namespace,
+      $object-level,
+      $classname-root,
+    )
+  );
+
+  .#{$classname}#{$breakpoint-suffix} {
+    @content;
+  }
+}
```
andreasknoepfle, line 142: This seems to be pretty useful :) 
I wonder why it is only used here... 
Do you want to introduce this in other places as well?

andreasknoepfle (approved):

scss/bitstyles/tools/_directional-properties.scss
planktonic: Yes exactly — I’ve already added generators for the directional utility classes (margin-right/-left etc), and the non-directional (e.g. grid-cols), this is the last one(?), that outputs a block of CSS in a standardised classname. Now that this exists, I would start using it throughout so we can provide every class at different breakpoints in a consistent manner

scss/bitstyles/atoms/card/settings.scss
planktonic: Yes that would totally work (it’s correct Sass too 🎉:D). But maybe it’s simpler to just output the `base` version outside the loop, wdyt?

```scss
@include output-block(
  $object-level: 'a',
  $classname-root: 'card'
) {
  @include card;
}

@each $breakpoint-alias in $bitstyles-card-breakpoints {
  @include media-query($breakpoint-alias) {
    @include output-block(
      $object-level: 'a',
      $classname-root: 'card',
      $breakpoint-suffix: '\\\@#{$breakpoint-alias}'
    ) {
      @include card;
    }
  }
}
```

The intent to output base then loop over breakpoints seems clearer, and it gets rid of the `if` statement so the rest of the loop is more readable too, I think

andreasknoepfle: Yes, that is even better 💯

planktonic: Perfect, done! 👍

PR #525, opened 2021-07-12
[#440] Select styles break without postcss?

### LINKED ISSUE #440: select styles break without postcss?
We use sass with postcss (autoprefixer, cssnano) on top. It seems that until postcss gets its hands on the CSS, the select styles are broken (likely due to the interpolated colors set on svg fills).

Investigate how we can avoid that, so the build system for consuming projects is more agnostic

planktonic:
Fixes #440 

- Adds postcss-loader so we can use autoprefixer (it should be included in every project for browser-compatibility)
- Adds a function to tools.color to URL-encode hex colors
- Uses it to fix the URL-encoded SVGs used for the `<select>` arrows, and for checkboxes’ ✔
- All our color functions now use namespaced sass functions

Now even when post-css is not being used to build bitstyles, the select icons are visible, so the bug is fixed.

⚠️ without post-css (and therefore autoprefixer), the `appearance` property isn’t prefixed to `-webkit-appearance`, and therefore the browser’s built-in icon and ui-chrome may be visible next to ours, as shown here:

|Before|After|
|--|--|
|before|after|

Delete if not applicable:

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -34,6 +34,16 @@ module.exports = {
         }
       },
       {
+        loader: "postcss-loader",
+        options: {
+          postcssOptions: {
+            plugins: [
+              ["autoprefixer"],
+            ],
+          },
+        }
+      },
+      {
         loader: "sass-loader"
       }],
       include: path.resolve(__dirname, '../scss'),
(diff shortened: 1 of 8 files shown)


andreasknoepfle (approved):
👏  awesome changes 👍

PR #530, opened 2021-07-19
Testing after migrating to `@use`

planktonic:
This is the basic testing structure I used while refactoring to use Sass’ `@use`/`@forward` module system. It stores fixtures of bitstyles CSS output with and without overrides, and `diff` s the current output of the branch against them (with no overrides, and with the same set of overrides). For intended changes, the branch should contain updates to these fixtures

Every method of `@import`/`@use`ing bitstyles gets tested. There’s an override for every module (so every class, sort of) that has variables, so we’ll have to add overrides when we add future modules)

❓Is using fixtures a reasonable approach? The horrible npm `test` script could be a bash script?

- Adds very long `test` script to package.json
- Adds fixtures for base bitstyles output, and bitstyles-with-overrides


package.json
```diff
@@ -9,6 +9,7 @@
     "lint:scss": "stylelint scss/**/*.scss --syntax scss",
     "lint:js": "eslint --ignore-path .gitignore --ext js,jsx .",
     "lint": "yarn lint:scss && yarn lint:js",
+    "test": "sass test/scss:test/scss/tmp --style=compressed --no-source-map && sass scss/bitstyles.scss:test/scss/tmp/bitstyles.css --style=compressed --no-source-map && postcss test/scss/tmp/*.css --replace && npx prettier --write test/scss/tmp/*.css && diff --from-file test/scss/fixtures/bitstyles-overrides.css test/scss/tmp/test-import-all.css && diff --from-file test/scss/fixtures/bitstyles-overrides.css test/scss/tmp/test-use-all.css && diff --from-file test/scss/fixtures/bitstyles-overrides.css test/scss/tmp/test-import-each.css && diff --from-file test/scss/fixtures/bitstyles.css test/scss/tmp/bitstyles.css",
```
klappradla, line 12: Yes, let's turn this into a bash script. Easier to format and therefore easier to read? Do you need help/inspiration for how to do it? I can also take over if you feel it's taking you too long.

Inspiration can be found at e.g. https://github.com/bitcrowd/tickety-tick

klappradla (commented):
Yay for tests 🎉 

I'm not sure I fully got the setup though 🙈. So we have a set of fixtures and we diff the output of the current state against those?

So whenever I change something, I'd also have to update the fixtures? _(similar to snapshot testing)_. How did you generate the fixtures?

Should we also run those tests on CI? They seem to be a good indicator for when dependency updates accidentally changed something 🤔

package.json
planktonic: Thanks! I’ll take a look over there 🙏

planktonic:
@klappradla thanks so much! Yes, when these test are run, the four manifest files are used to compile the Sass. Each of the four manifest files uses a different method for overriding variables, either `@import`s or `@use`s each of the bitstyles modules, and either imports the whole library in one line, or imports each module one by one.

The goal is to test that overriding the variables works for each of the possible methods of consuming bitstyles — since the switch to Sass modules (`@use`), we’re supporting the old way `@import`  in order to remain backwards-compatible. So there’s one fixture file, all the test manifests change the same variables to the same values, and the output of all of them should match the fixture.

Yes, a side-effect is that this is like snapshot testing for the whole library too. So any intentional changes to e.g. classnames or default values for variables, which will result in different CSS output, would need to be propagated to the fixture too.

I’ll update to use a bash script here, and add the CI testing in another PR. Issue: #547

klappradla:
> Yes, a side-effect is that this is like snapshot testing for the whole library too. So any intentional changes to e.g. classnames or default values for variables, which will result in different CSS output, would need to be propagated to the fixture too.

Hm, could be over-engineering it, but for the sake of maintenance it could still make sense: how about adding Jest and writing the tests as JS-snapshot tests? From what I remember, it does not really matter "how" the content of the snapshot is generated, so one could probably easily shell out to Sass for doing it. With the snapshots in place, diffing and updating could become a lot easier.

What do you think?

If it's a super wild idea, @pmeinhardt & I can probably also take a look at it during one of our Hacknights 🦉

package.json
planktonic: I’ve updated, this is now handled in a bash script (called via `yarn test` for convenience)

planktonic:
@klappradla I updated to use a bash script, but left the functionality the same

If you think a Jest setup would be better (the current setup is quite hand-made I guess, an established snapshot system is very likely better), and if you have the time, that would be great! If you decide to, let me know what I can do to help

script/test
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+
+set -o errexit
+set -o nounset
+
+trap cleanup 1 2 15
```
klappradla, line 6: Never seen `trap` 💡

script/test
```diff
@@ -0,0 +1,32 @@
...
+
+cleanup() {
+  rm -rf "$temp_folder"
+}
+
+temp_folder="./test/tmp"
+fixtures_folder="./test/scss/fixtures"
```
klappradla, line 13: You should probably define those above `cleanup`. And I guess it's best practice to write them in all caps.

script/test
```diff
@@ -0,0 +1,32 @@
...
+npx sass scss/bitstyles.scss:$temp_folder/bitstyles.css --style=compressed --no-source-map
+npx postcss $temp_folder/*.css --replace
+npx prettier --write $temp_folder/*.css
+
+echo "\nComparing generated CSS against expected CSS…\n"
+
+diff -u -w --from-file $fixtures_folder/bitstyles-overrides.css $temp_folder/test-import-all.css \
+&& diff -u -w --from-file $fixtures_folder/bitstyles-overrides.css $temp_folder/test-use-all.css \
+&& diff -u -w --from-file $fixtures_folder/bitstyles-overrides.css $temp_folder/test-import-each.css \
+&& diff -u -w --from-file $fixtures_folder/bitstyles.css $temp_folder/bitstyles.css
```
klappradla, line 27: I'd intend those one level deeper for the sake of readability.

script/test
```diff
@@ -0,0 +1,32 @@
...
+
+compare_result=$?
+
+cleanup
+exit $compare_result
```
klappradla, line 32: Wonder how actionable this is for developers 🤔

In order to know "what"  is off, I'd still have to run the internals of script myself, right? Maybe it makes sense to output the actual diff result as well.

Or am I missing something?

klappradla (changes requested):
Neat 👍 

Maybe briefly run the script through https://github.com/koalaman/shellcheck to see if there are some best practices violated.

I'm also not too sure about the output of the `test` command: if it's just a simple "yes" or "no", people may still need to dive into the internals of the script and run them in isolation in order to figure out what's wrong. If so, we should give them more actionable information in my opinion.

script/test
planktonic: The diff commands further up output any differences they find to stdout, so if there’s a problem it is actionable (as much as the diff output is readable. I tried to pick a good output format, but possibly YMMV)

planktonic:
@klappradla thanks for taking another look 🎉 I updated the test script, and ran it through the shellchecker. See comments about script output when a diff is found — the output goes to stdout, should be enough I think?

script/test
klappradla: **TIL** "YMMV" 🤯 

https://www.howtogeek.com/693183/what-does-ymmv-mean-and-how-do-you-use-it/

klappradla (approved):

script/test
planktonic: :D am I old, or is it because it’s English? 🙈

klappradla:
> Hm, could be over-engineering it, but for the sake of maintenance it could still make sense: how about adding Jest and writing the tests as JS-snapshot tests? From what I remember, it does not really matter "how" the content of the snapshot is generated, so one could probably easily shell out to Sass for doing it. With the snapshots in place, diffing and updating could become a lot easier.

Quick update on this: Paul & I thought this through yesterday night and decided against a Jest & snapshots setup. Main point against it was: Jest only ever allows to update snapshots as a whole, which can get quite hard on the eyes for files as large as the fixtures for these tests right here. Therefore, you're probably for now still better off with `git add --patch` or something like this.

PR #537, opened 2021-07-29
[#536] Linting

### LINKED ISSUE #536: Linting
Some variables do not have `!default`, meaning they cannot be overridden by users. We should add linting (*possibly the [stylelint-scss](https://github.com/kristerkari/stylelint-scss) stylelint plugin?), and fix all those that it finds

planktonic:
Fixes #536 

- Adds `stylelint-scss`, so we can lint the sass-specific syntax
- Fixes a few cases where variables were not declared with `!default`
- Disables the linting on the exceptions (all in `tools`, where we’re declaring a base value for a variable, then re-specifying it inside a conditional)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Fixed
+
+- All of the variables related to inputs, selects, and fieldsets can now be overridden by users
+
 # [[3.0.0-rc.1]](https://github.com/bitcrowd/bitstyles/releases/tag/v3.0.0-rc.1) - 2021-08-13
 
 ## Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Fixed
 
-- All of the variables related to inputs, selects, and fieldsets can now be overridden by users
+- All of the variables can now be overridden by users
 - Sizing of checkboxes & radios now works by setting the `$bitstyles-checkbox-size` and `$bitstyles-radio-size` variables respectively
 
 # [[3.0.0-rc.1]](https://github.com/bitcrowd/bitstyles/releases/tag/v3.0.0-rc.1) - 2021-08-13
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## Fixed
 
 - All of the variables can now be overridden by users
-- Sizing of checkboxes & radios now works by setting the `$bitstyles-checkbox-size` and `$bitstyles-radio-size` variables respectively
 
 # [[3.0.0-rc.1]](https://github.com/bitcrowd/bitstyles/releases/tag/v3.0.0-rc.1) - 2021-08-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Fixed
 
 - All of the variables related to inputs, selects, and fieldsets can now be overridden by users
+- Sizing of checkboxes & radios now works by setting the `$bitstyles-checkbox-size` and `$bitstyles-radio-size` variables respectively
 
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -81,6 +81,7 @@
     "style-loader": "^2.0.0",
     "stylelint": "^13.13.1",
     "stylelint-config-bitcrowd": "^4.0.0",
+    "stylelint-scss": "^3.20.1",
```
klappradla, line 84: Maybe mention in the README how to best run this (I guess it also adds an `npm` task)

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# Unreleased
+
+## Fixed
+
+- All of the variables can now be overridden by users
+- Sizing of checkboxes & radios now works by setting the `$bitstyles-checkbox-size` and `$bitstyles-radio-size` variables respectively
```
klappradla, line 6: Is this "breaking" 👀 ?

scss/bitstyles/base/input-checkbox/_settings.scss
```diff
@@ -6,7 +6,7 @@
 // Base styles ////////////////////////////////////////
 $border-radius: size.get('xxs') !default;
 $gap: size.get('xs') !default;
-$size: size.get('m') !default;
+$size: size.get('l') !default;
```
klappradla, line 9: Was this a bug?

scss/bitstyles/tools/_color.scss
```diff
@@ -15,10 +15,10 @@
 // @param {list} $percentages  A list of decmial values representing the percentages at which the mix-color will be mixed into the base-color
 
 @function generate-palette($base-color, $mix-color, $percentages) {
-  $color-map: ();
+  $color-map: (); // stylelint-disable-line scss/dollar-variable-default
```
klappradla, line 18: Hm, a bit weird that the linter does not get the function context here 🤔 

Does the fix from [this issue](https://github.com/kristerkari/stylelint-scss/issues/76) make a difference?

scss/bitstyles/tools/_directional-properties.scss
```diff
@@ -2,6 +2,8 @@
 @use 'sass:list';
 @use 'sass:string';
 
+/* stylelint-disable scss/dollar-variable-default */
```
klappradla, line 5: See my earlier comment: I think/hope stylelint has an option for this already 😉

klappradla (changes requested):
Yay for linting 💯 

I'd suggest to add a note about the linter in the contributors' section of the REAME. Plus: it probably makes sense to run it on CI.

The disabling of rules generates a lot of noise. Maybe you can quickly check if the outcomes of [this issue](https://github.com/kristerkari/stylelint-scss/issues/76) can make this easier for these cases?

package.json
planktonic: There’s no new npm script needed — this is a plugin for stylelint that adds some sass-related/useful rules to the default set provided by stylelint. So when you run `yarn lint`, these rules are used too

CHANGELOG.md
planktonic: Ahh, sorry this is something better factored out into its own PR — I‘ll do that!

scss/bitstyles/base/input-checkbox/_settings.scss
planktonic: Something I found along the way, which should be in its own PR (coming soon…)

package.json
planktonic: …there was no section about linting there though, so I added one ✨

scss/bitstyles/tools/_color.scss
planktonic: It did! I updated the linting rules, thanks 🎉

klappradla (approved):

PR #539, opened 2021-07-29
[#534] Changelog heading structure

### LINKED ISSUE #534: Changelog heading structure
The changelog should only have one top-level heading, that reads `Changelog`. All other headings should be shifted down one level

planktonic:
Fixes #534 

- Fixed the heading structure of the changelog


klappradla (approved):
Makes sense 👍

PR #555, opened 2021-11-10
[#554] Align dt/dd to baseline

### LINKED ISSUE #554: align dt/dd to baseline
The `<dt>` and `<dd>` are different font-sizes in our example code for `atoms/dl`, with the result that they do not align to the same baseline. This becomes more and more noticeable when there’s a larger difference in font-sizes. We could use `u-items-baseline` on the parent `<div>`

planktonic:
Fixes #554 

- Example `dl`s are now aligned to baseline

## Looks like

Screenshot 2021-11-10 at 16 40 55

Delete if not applicable:


andreasknoepfle (approved):
👍

PR #566, opened 2021-11-17
[#564] Remove aria-hidden for the alpine examples when display: none

### LINKED ISSUE #564: Remove `aria-hidden` for the alpine examples when `display: none`
When implementing the dropdown menu in bitstyles_phoenix I noticed, that we are setting `aria-hidden` in the alpine examples for things like hidden menus in the dropdown. But this page suggests that it is best practice to not set them when there is already `display: none` set (which alpine will already set). 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute#best_practices

I decided to not set them in `bitstyles_phoenix` and we can maybe remove them here as well in order not to confuse anyone?

planktonic:
Fixes #564 

- Removes all instances of `aria-hidden` from the dropdowns used in all the example code, as it was already handled by the show/hide (display block/none) applied in JS


andreasknoepfle (approved):
👏

PR #568, opened 2021-11-17
[#559] Allow font base url to be configured

### LINKED ISSUE #559: Allow font base url to be configured
At the moment, the `../assets/fonts` url in the settings are hardcoded. This works great for any kind of sass processors that supports copying files to the output (aka webpack & postcss with file loader) but breaks a bit for loaders that simply ignore `url` properties and spit them out as is (dart-sass standalone). 
So it would be great to have a config variable for the path with the current value as default.

https://github.com/bitcrowd/bitstyles/blob/main/scss/bitstyles/settings/_typography.scss

planktonic:
Fixes #559 

- Font-file location is now set by a variable
- Updates tests to use a different  webfont name instead of a standard `sans-serif`

Delete if not applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changed
 
 - `dl` examples are now aligned to the baseline, using the `u-items-baseline` class. Update `dl` classnames in your project to match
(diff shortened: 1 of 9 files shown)


andreasknoepfle (approved):
🤩

PR #570, opened 2021-11-18
[#561] Add variable to handle margins in base/typography

### LINKED ISSUE #561: Add variable to handle margins in base/typography
Default of no margin is good, but it’d be great to be able to override that

planktonic:
Fixes #561 

- Default margins for heading elements and paragraphs can now be specified with the `$margin-heading` and `$margin-paragraph` variables respectively
- The default margin for paragraphs is now `0`

Delete if not applicable:

### DIFF
--- a/scss/bitstyles/generic/_normalize.scss
+++ b/scss/bitstyles/generic/_normalize.scss
@@ -171,14 +171,6 @@ button:-moz-focusring,
 }
 
 /**
- * Correct the padding in Firefox.
- */
-
-fieldset {
-  padding: size.get('m');
-}
-
-/**
  * 1. Correct the text wrapping in Edge and IE.
  * 2. Correct the color inheritance from `fieldset` elements in IE.
  * 3. Remove the padding so developers are not caught out when they zero out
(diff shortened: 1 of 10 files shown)


scss/bitstyles/generic/_normalize.scss
```diff
@@ -170,14 +170,6 @@ button:-moz-focusring,
   outline: 1px dotted ButtonText;
 }
 
-/**
- * Correct the padding in Firefox.
- */
-
-fieldset {
-  padding: size.get('m');
```
maltoe, line 178: drive-by?

maltoe (approved):
LGTM 👍

scss/bitstyles/generic/_normalize.scss
planktonic: Yes kinda :) It’s related to the issue, in that it’s about default spacing applied to elements at this early stage, which is why I found it. This property gets re-specified later in the forms file (using a variable that can be overridden in projects, which for typography was the purpose of the PR), so this line was not needed

PR #577, opened 2021-12-03
add u-gap-xs

maltoe:
👋


planktonic (commented):
Interesting! What’s the use case, is it a general enough need to be made default?

maltoe:
> Interesting! What’s the use case, is it a general enough need to be made default?

I don't know. Thought it was more of an "omission". use-case is the buttons in the project-whose-name-we-don't-say, in the tables. Found the `s` gap a bit too wide, `xs` also gives more room for actual content in the table.

planktonic:
Ok yeh that’s reasonable :+1: I’ll take a look again and look to get this merged

planktonic (approved):
Thanks for the addition! ✨

PR #581, opened 2022-01-03
[#576] select remove half the border width from vertical padding

### LINKED ISSUE #576: select: Remove half the border width from vertical padding
Hey @planktonic 

Today I learned something about `<select>` and `<input>` elements in one of our projects (you know which):

The current settings make them both equal in height, but they have different borders on `:focus`:

| input focussed | select focussed |
| - | - |
|Screenshot 2021-12-02 at 22 17 17|Screenshot 2021-12-02 at 22 17 23|

Reason for this is that in our custom settings we have defined

```sass
$padding: size.get('s'),
$border-width: size.get('xxs'),
// among other settings...
```
 
on the text input but not on the `select`:

```sass
$padding-vertical: size.get('s'),
// missing `border-width`
```

If I add that definition, the border is fine, but the `<select>` element becomes taller than its `<input>` friends

Screenshot 2021-12-02 at 22 17 41

Reason for that is that for `<input>` you deduct half the border width from the vertical padding:

https://github.com/bitcrowd/bitstyles/blob/d4d94a1c2f5fba31ef73b6e5d525cb92b1190d52/scss/bitstyles/base/input-text/_input-text.scss#L41

... while for `<select>`s you don't

https://github.com/bitcrowd/bitstyles/blob/d4d94a1c2f5fba31ef73b6e5d525cb92b1190d52/scss/bitstyles/base/select/_select.scss#L10

So essentially setting `$vertical-padding` in our project like this:

```sass
$padding-vertical: (size.get('s') - math.div(size.get('xxs'), 2)),
```

Soooo. I'm fine with just setting it in this project, but I wonder if this is a 🐛  in bitstyles and should be fixed? OTOH, it would likely break existing designs, so maybe that is also not an option?

Either way, I thought I'd mention it to you.

👋

planktonic:
Fixes #576

- Inputs, selects, and buttons all have same border/line-height/padding settings & CSS, so they always match height
- Adds variables for font styles, and makes the padding variables all match the same pattern

Delete if not applicable:

- [ ] Storybook documentation has been updated
- [ ] Your changes have been added to the `unreleased` section of [CHANGELOG.md](../CHANGELOG.md)

### DIFF
--- a/scss/bitstyles/atoms/button/_button.scss
+++ b/scss/bitstyles/atoms/button/_button.scss
@@ -1,7 +1,6 @@
 @use 'sass:math';
 @use './settings';
 @use '../../settings/setup';
-@use '../../settings/typography';
 
 .#{setup.$namespace}a-button {
   position: relative;
@@ -14,9 +13,8 @@
   padding: settings.$padding-vertical settings.$padding-horizontal;
   margin: 0;
   overflow: visible;
-  font: inherit;
-  font-weight: typography.$font-weight-normal;
-  line-height: typography.$line-height-min;
+  font: settings.$font;
+  font-weight: settings.$font-weight;
   text-align: center;
   text-decoration: none;
   white-space: nowrap;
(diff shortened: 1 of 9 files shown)


maltoe (approved):
✈️  LGTM as far as I can tell

PR #589, opened 2022-01-04
[#586] variables to set color background color of html

### LINKED ISSUE #586: Variables to set color, background-color of html
For cases where the Color-palette has been overridden, there may not be a `text` color, nor `background` color. But right now `base/typography` uses grabs those colors directly from `palette.get()`. Change to using variables to specify them both, so we can override that in projects

planktonic:
Fixes #586 

- `html` styles are now split out from the typography base styles
- color & background color on the `html` element can now be specified with sass variables

Delete if not applicable:

### DIFF
--- a/scss/bitstyles/base/html/html.stories.mdx
+++ b/scss/bitstyles/base/html/html.stories.mdx
@@ -1,0 +1,16 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
+
+<Meta title="Base/Html" />
+
+# HTML
+
+Applies some basic styles to the HTML element, to cascade down the DOM.
+
+## Customization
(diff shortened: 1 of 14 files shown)


scss/bitstyles/base/html/html.stories.mdx
```diff
@@ -0,0 +1,16 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
```
klappradla, line 1: Nice ✨ 

Didn't know this was possible.

On that note: does it make sense to apply a certain order to the sidebar in Storybook?

klappradla (approved):
Nice improvement - good idea 👍

PR #590, opened 2022-01-04
[#587] variable to specify default value for palette get

### LINKED ISSUE #587: variable to specify default value for palette.get()
Right now, `palette.get()` will return the color passed, defaulting to the value at `"100"`, as that’s our default color palette scale. For Color-palettes that use different naming, it’d be great to be able to specify a different key name to take when none is passed.

planktonic:
Fixes #587 

- Default key-name to return when asking for a color without specifying it (e.g. `palette.get('white')`) can now be set using `$bitstyles-color-palette-default-key`

Delete if not applicable:

### DIFF
--- a/scss/bitstyles/settings/_color-palette.scss
+++ b/scss/bitstyles/settings/_color-palette.scss
@@ -109,3 +109,5 @@ $color-palette: (
     $percentages: $color-mix-percentages
   ),
 ) !default;
+
+$default-key: '100' !default;
--- a/scss/bitstyles/tools/_palette.scss
+++ b/scss/bitstyles/tools/_palette.scss
@@ -8,19 +8,19 @@
 //
 // @param {String} $color  Base colour within the palette
 //
-// @param {String} $percent   Optional. Percentage of the color to mix with white. Defaults to `100%` (fuly the specified color, no white)
+// @param {String} $key   Optional. Percentage of the color to mix with white. Defaults to `100%` (fuly the specified color, no white)
 //
 
-@function get($color, $percent: '100') {
+@function get($color, $key: color-palette.$default-key) {
   @if map.has-key(color-palette.$color-palette, $color) {
     $palette: map.get(color-palette.$color-palette, $color);
 
-    @if map.has-key($palette, $percent) {
-      @return map.get($palette, $percent);
+    @if map.has-key($palette, $key) {
+      @return map.get($palette, $key);
     }
 
     @else {
-      @error 'Oops! Color/percentage ‘#{$color}/#{$percent}’ does not exist.';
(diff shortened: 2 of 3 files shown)


scss/bitstyles/tools/_palette.scss
```diff
@@ -8,19 +8,19 @@
 //
 // @param {String} $color  Base colour within the palette
 //
-// @param {String} $percent   Optional. Percentage of the color to mix with white. Defaults to `100%` (fuly the specified color, no white)
+// @param {String} $key   Optional. Percentage of the color to mix with white. Defaults to `100%` (fuly the specified color, no white)
```
klappradla, line 11: The description still applies to "percentage" though.

scss/bitstyles/settings/_color-palette.scss
```diff
@@ -109,3 +109,5 @@ $color-palette: (
     $percentages: $color-mix-percentages
   ),
 ) !default;
+
+$default-key: '100' !default;
```
klappradla, line 113: Just wondering: doesn't this increase the complexity of setting up bitstyles in a project quite a bit? I could imagine, that for the special cases, where the keys don't follow the classic scheme, one would instead require to just use the second parameter.

The way I understand it, the function already explodes when the color and percentage pair does not exist. So we wouldn't run into the problem of us accidentally not specifying it for these projects.

Just a though, maybe I'm also not fully getting the problem this PR is supposed to fix.

klappradla (commented):
Valid 👍 

However, I think I don't fully get the problem or the "severeness" of the issue this is supposed to solve 😉 

In my opinion, the problem this is trying to solve is so small, that the price in terms of complexity, which another settings switch adds, is quite high.

scss/bitstyles/settings/_color-palette.scss
planktonic: When setting up, there’s no extra complexity if you’re suing the default color palette, or another palette with the same naming of colors. This change makes it more convenient to use the `palette.get()` function if you have a different naming scheme.

This makes it easier to use naming like:

```scss
'gray': (
  'default': #999,
  'light': #ccc,
  'dark': #666,
)
```
…because you can get the 'default' color by default when using `palette.get('gray')`.

Without this change, the function will throw an error on `palette.get('gray')`, because it looks for the `100` key, which doesn’t exist.

As you say, projects can explicitly use `palette.get('gray', 'default')`, but the convenience seemed worth the extra variable. Does that make sense, and does it seem worth it to you?

klappradla: I see! Thanks for the explanation, the use case just wasn't obvious for me at first 💡 

Makes perfect sense! Nice addition to bitstyles 💖

klappradla (approved):
🔥

PR #592, opened 2022-01-05
[#588] allow sidebar organism to change sides

### LINKED ISSUE #588: Allow Sidebar organism to change sides
Right now the sidebar is hardcoded to render on the left. This should be overridable in settings, so it renders either on the left, or on the right (perhaps follow the pattern set by dropdowns for handling directionality)

planktonic:
Fixes #588

- Sidebar organism can be positioned on the left or right, specified using a variable

Delete if not applicable:

### DIFF
--- a/scss/bitstyles/organisms/sidebar/_sidebar.scss
+++ b/scss/bitstyles/organisms/sidebar/_sidebar.scss
@@ -11,7 +11,7 @@
   z-index: 1;
   top: 0;
   bottom: 0;
-  left: 0;
+  #{settings.$small-location}: 0;
   width: settings.$small-width;
   box-shadow: settings.$small-box-shadow;
 
--- a/scss/bitstyles/organisms/sidebar/_sidebar.scss
+++ b/scss/bitstyles/organisms/sidebar/_sidebar.scss
@@ -6,6 +6,7 @@
   width: settings.$large-width;
 }
 
+/* stylelint-disable declaration-empty-line-before */
 .#{setup.$namespace}o-sidebar--small {
   position: fixed;
   z-index: 1;
@@ -32,3 +33,4 @@
     }
   }
 }
+/* stylelint-enable declaration-empty-line-before */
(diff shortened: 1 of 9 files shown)


scss/bitstyles/organisms/sidebar/_sidebar.scss
```diff
@@ -6,12 +6,13 @@
   width: settings.$large-width;
 }
 
+/* stylelint-disable declaration-empty-line-before */
```
klappradla, line 9: Given the size of the block here, maybe we're better of with "next-line" ignoring or inline ignoring. See https://stylelint.io/user-guide/ignore-code/

Just a thought.

scss/bitstyles/ui/sidebar.stories.mdx
```diff
@@ -346,3 +346,38 @@ Uses a fullwidth dropdown to fit with no overlap.
...
+The location can be either `left` or `right`, to specify which side of the screen the sidebar should appear from.
+
+```scss
+$small-location: left;
+$small-box-shadow: size.get('m') 0 size.get('l') rgba(palette.get('brand-1', '100'), 0.1),
+  size.get('l') 0 size.get('m') rgba(palette.get('gray', '90'), 0.07),
+  0 0 size.get('xl') rgba(palette.get('gray', '90'), 0.1);
+```
+
+By default the sidebar will animate into view and back out, assuming the classes `.is-off-screen` and `.is-on-screen` will be applied using JS. The classnames and the styles to be applied can be overridden. Be sure to update the `$will-change` to reflect the property you’re transitioning. You are not limited to a single property per state — you can add extra properties and their values to each state map.
```
klappradla, line 370:
```suggestion
By default the sidebar will animate into view and back out, assuming the classes `.is-off-screen` and `.is-on-screen` will be applied using JS. The classnames and the styles to be applied can be overridden. Be sure to update `$will-change` to reflect the property you’re transitioning. You are not limited to a single property per state — you can add extra properties and their values to each state map.
```

klappradla (approved):
Neat!

PR #594, opened 2022-01-06
[#585] get() for font sizes

### LINKED ISSUE #585: .get() for font-sizes
We have helper functions to get colors and sizes, but no equivalent for font-sizes (`font-size.get()`?). It would be great to handle breakpoints with this, but it’s also fine to require the breakpoints be handled at point of use (and for the function to therefore return font-size for a given breakpoint)

planktonic:
Fixes #585 

- Adds a helper function `font-size.get()` that returns a given font-size
- Complexity is removed from the typography tools which generate the element- and class-based font sizes — the helper function is used there instead
- Renames the media-query mixin to match the other getters

Delete if not applicable:


CHANGELOG.md
```diff
@@ -9,21 +9,20 @@
...
-
 - Modals now only show scrollbars (on OSes/configs where scrollbars are shown) when they’re needed
 
 ### Breaking
 
 - All utility classes which had a double dash in their classname (`--`) now only have a single dash. You’ll need to rename all these classes to use single dashes e.g. `.u-bg--brand-1` becomes `.u-bg-brand-1`
 - Values for the display classes (`u-block`, `u-flex` etc.) will now need to be set unquoted. If you previously were overriding the `$bitstyles-display-values` variable, make sure all the values on the right hand side are unquoted CSS values e.g. `'block': block` instead of `'block': 'block'`, as was previously possible
-- If you were using the `$bitstyles-input-padding` variable, you’ll now need to set `$bitstyles-input-padding-vertical` and `$bitstyles-input-padding-horizontal` spearately
+- If you were using the `$bitstyles-input-padding` variable, you’ll now need to set `$bitstyles-input-padding-vertical` and `$bitstyles-input-padding-horizontal` separately
+- The media-query/breakpoint mixin is no longer called using `media-query.media-query()`. You’ll need to update all uses of this to `media-query.get()`
```
klappradla, line 25: 🙌

scss/bitstyles/tools/_font-size.scss
```diff
@@ -0,0 +1,42 @@
+@use 'sass:map';
+@use '../settings/typography';
+@use 'unit-conversions';
+
+// Palette
```
klappradla, line 5: 👀 don't really see how this relates to the file name or function 🤔

scss/bitstyles/tools/_font-size.scss
```diff
@@ -0,0 +1,42 @@
...
+      $base-size: map.get($font-sizes, 'html');
+    }
+
+    @if map.has-key($font-sizes, $font-size) {
+      @if $font-size == 'html' {
+        @return unit-conversions.px-to-rem(map.get($font-sizes, $font-size), 16px);
+      }
+
+      @else {
+        @return unit-conversions.px-to-rem(map.get($font-sizes, $font-size), $base-size);
```
klappradla, line 30: Is there a guarantee that the `font-size-base-small` is defined as `px`? Or what happens with this conversion if it's not in px?

scss/bitstyles/tools/_font-size.scss
```diff
@@ -0,0 +1,42 @@
...
+    $font-sizes: map.get(typography.$font-sizes, $breakpoint);
+    $base-size: typography.$font-size-base-small;
+
+    @if map.has-key($font-sizes, 'html') {
+      $base-size: map.get($font-sizes, 'html');
+    }
+
+    @if map.has-key($font-sizes, $font-size) {
+      @if $font-size == 'html' {
+        @return unit-conversions.px-to-rem(map.get($font-sizes, $font-size), 16px);
```
klappradla, line 26: Where do the hardcoded 16px come from?

CHANGELOG.md
```diff
@@ -9,21 +9,20 @@
...
-### Fixed
-
 - Modals now only show scrollbars (on OSes/configs where scrollbars are shown) when they’re needed
 
 ### Breaking
 
 - All utility classes which had a double dash in their classname (`--`) now only have a single dash. You’ll need to rename all these classes to use single dashes e.g. `.u-bg--brand-1` becomes `.u-bg-brand-1`
 - Values for the display classes (`u-block`, `u-flex` etc.) will now need to be set unquoted. If you previously were overriding the `$bitstyles-display-values` variable, make sure all the values on the right hand side are unquoted CSS values e.g. `'block': block` instead of `'block': 'block'`, as was previously possible
-- If you were using the `$bitstyles-input-padding` variable, you’ll now need to set `$bitstyles-input-padding-vertical` and `$bitstyles-input-padding-horizontal` spearately
+- If you were using the `$bitstyles-input-padding` variable, you’ll now need to set `$bitstyles-input-padding-vertical` and `$bitstyles-input-padding-horizontal` separately
```
klappradla, line 24: 🧑‍🏫

klappradla (approved):
Nice cleanup 👏 

The newly added function looks a bit complex for what it does though 👀

CHANGELOG.md
planktonic: ⚔️🛡

scss/bitstyles/tools/_font-size.scss
planktonic: That’s the standard browser default font-size. I could make it a variable, so the name would make that clearer

scss/bitstyles/tools/_font-size.scss
planktonic: No, there’s no guarantee. And it’s not so easy to check unit type in sass (there’s a built-in function but it’s not guaranteed to output the same result in different platforms & sass implementations :/ )

We could require that the font-sizes are actually given in `rem` in the first place, wdyt?

Rems are pretty well-understood by devs by now, other libraries use them directly. It would mean we can remove this conversion totally. And it would remove the need for the other issue you raised here, about the 16px value - that’s only there for the conversion to rem 🤔

scss/bitstyles/tools/_font-size.scss
planktonic: Yup, thanks! Copy/paste error

scss/bitstyles/tools/_font-size.scss
planktonic: I updated — fixed both problems you raised by specifying the font-sizes directly in rem 🧹

klappradla: Also an option. Wondering if that makes the upgrade path harder for existing projects though 🤔

planktonic: This isn’t a breaking change, existing projects could continue using `px` to specify font-sizes. Ideally they’d switch to using `rem` though, as that’s better CSS for the end user, so yes this adds a ‘thing to check’ to the list when updating. I think it’s not a big change tho, and the next release is a major version, so I’d go with it

PR #607, opened 2022-01-24
[#595] Add utility: box-shadow

planktonic:
Fixes #595 

The following changes are contained in this pull request:
- adds helper functions for getting & generating shadows
- adds drop-shadow & box-shadow utility classes
- Uses the shadow generator for all the existing box-shadows we use on other components

Delete if not applicable:

- [x] Storybook documentation has been updated
- [x] Your changes have been added to the `unreleased` section of [CHANGELOG.md](../CHANGELOG.md)

PR #609, opened 2022-01-25
[#604] Update linting

### LINKED ISSUE #604: Update linting
Update stylelint and the linting config

planktonic:
Fixes #604 

- Updates to stylelint v14
- Updates to stylelint-config-bitcrowd v5.0.1
- Runs prettier, and applies linting


planktonic:
TODO:

- [ ] See https://github.com/bitcrowd/bitstyles/pull/607#discussion_r792427622. Check the table format after that’s merged, and this is rebased
- [ ] Update the linting config at source, and then update here (should mean no src changes, only moving linting rules to the most appropriate place)

scss/bitstyles/settings/shadows.stories.mdx
```diff
@@ -33,12 +33,12 @@ This creates a specification for a multi-layered shadow using the `brand-1` colo
 
 ### Parameters
 
-|Name|Format|Description|
-|--|--|--|
-|$color|color|The color to be used. Probably you want a low-opacity version of a color. Avoid using pure black unless your design calls for it, as that can look flat, especially on non-white backgrounds.|
-|$offset|number|Vertical offset for the shadow. Include the units unless it’s zero. Optional, defaults to `0`.|
-|$blur-radius|number|Maximum distance to blur the shadow. The extra layers in this multi-layer shadow will be smaller than the value you pass. Include the units.|
-|$layers|number|Number of layers of in the shadow. Optional, defaults to `4`.|
+| Name         | Format | Description                                                                                                                                                                                   |
```
planktonic, line 36: @klappradla prettier handled the tables well I think! Except no line-wrapping

.travis.yml
```diff
@@ -9,7 +9,7 @@ before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash
   - export PATH=$HOME/.yarn/bin:$PATH
 node_js:
-  - "12.13.0"
+  - '12.13.0'
```
klappradla, line 12: Hm, this is a completely different version than the one in `.tool-versions`. Aside that: does the project even still run on Travis (👴). Didn't they shut down?

CHANGELOG.md
```diff
@@ -84,7 +84,7 @@
 
 ### Added
 
-- The entire library is now compatible with the [Sass module system](https://sass-lang.com/blog/the-module-system-is-launched), and backwards compatibile with the existing Sass import system. This means you can import  the whole library or its parts with `@use`/`@forward` or `@import`, and override variables in all of the available ways. The functions and mixins are all available in both systems
+- The entire library is now compatible with the [Sass module system](https://sass-lang.com/blog/the-module-system-is-launched), and backwards compatibile with the existing Sass import system. This means you can import the whole library or its parts with `@use`/`@forward` or `@import`, and override variables in all of the available ways. The functions and mixins are all available in both systems
```
klappradla, line 87: How did you run prettier on those `.md` doc files? Does the project's `lint` task automatically include those? Otherwise we should add a custom task for it to prevent them from going off again in the future 🤔 

I can take care of the extra setup if we don't have it. Just assign me an issue.

package.json
```diff
@@ -6,18 +6,18 @@
   "scripts": {
     "build": "sass scss/bitstyles.scss build/bitstyles.css --style=compressed --no-source-map",
     "postbuild": "postcss build/bitstyles.css --replace",
-    "lint:scss": "stylelint scss/**/*.scss --syntax scss",
+    "lint:scss": "stylelint --ignore-path .gitignore 'scss/**/*.scss' --customSyntax postcss-scss",
     "lint:js": "eslint --ignore-path .gitignore --ext js,jsx .",
     "lint": "yarn lint:scss && yarn lint:js",
+    "prettier": "prettier --ignore-path .gitignore --check ./scss",
```
klappradla, line 12: Is this also taking care of `.md` files? Should we add a separate `lint:md` for it?

Aside that: wouldn't it be easier for devs to handle, if we'd include prettier in the eslint config and then have `lint:js` already take care of this? Same for stylelint of course.

It's all possible with their plugin systems: https://prettier.io/docs/en/integrating-with-linters.html

klappradla (commented):
Haven't looked through all the files yet, but here's already some thoughts about the general setup 🤔 

Most importantly though: isn't Travisci.org shut down? Do we even have CI for this project 😱 ?

planktonic:
@klappradla amazing, thanks for the feedback! I’ll take a look and update 💃

Yes, sadly there are a lot of changes, as the linting (especially the ordering! :D) are now slightly different.

CHANGELOG.md
planktonic: Yes, I’ll add that to the package.json & CI 🙏

package.json
planktonic: They suggest using the plugins is not ideal, and using `prettier` on the CLI directly is better. Since our previous projects use `prettier` directly, I stuck with that approach, but added prettier tasks for js, scss, and mdx. Those run on each lint task, and therefore on CI. If you think it’s better we can switch to using plugins though, wdyt?

klappradla: No strong feelings here. Fine to have them as separate tasks, as long as they're discoverable and run on CI.

klappradla:
It looks like the file paths for the linter tasks aren't dialed in correctly yet. At least it's 🚨 on CI. Seems like it's looking into the `node_modules` directory.

planktonic:
@klappradla yup, it was passing locally, but seems it fails on CI when the paths aren’t quoted  🙈 Fixed!

package.json
```diff
@@ -6,18 +6,20 @@
   "scripts": {
     "build": "sass scss/bitstyles.scss build/bitstyles.css --style=compressed --no-source-map",
     "postbuild": "postcss build/bitstyles.css --replace",
-    "lint:scss": "stylelint scss/**/*.scss --syntax scss",
-    "lint:js": "eslint --ignore-path .gitignore --ext js,jsx .",
-    "lint": "yarn lint:scss && yarn lint:js",
+    "lint:scss": "stylelint --ignore-path .gitignore 'scss/**/*.scss' --customSyntax postcss-scss && yarn prettier:scss",
+    "lint:js": "eslint --ignore-path .gitignore --ext js,jsx . && yarn prettier:js",
+    "lint:mdx": "prettier --ignore-path .gitignore --check './**/*.mdx'",
```
klappradla, line 11: Hm, I guess we're missing the `.md` files with that and you already ran prettier on them. Ignoring them from now on could mean they go off the rails.

How about:

```suggestion
    "lint:docs": "prettier --ignore-path .gitignore --check './**/*.{md,mdx}'",
```

planktonic: 👍 done (needed another change where this task was used in another task)

klappradla (approved):
👨‍🎤

PR #611, opened 2022-01-27
[#591] button generator

### LINKED ISSUE #573: Dangerous icon buttons
Dangerous icon buttons are currently not possible, since the hover from the `a-button--icon` overrides the `danger`.

```
<a class="a-button a-button--icon a-button--danger foo" href="#" title="Delete">
  <svg xmlns="http://www.w3.org/2000/svg" aria-hidden="true" class="a-icon" focusable="false" height="16" width="16">
    <use xlink:href="assets/icons.svg#icon-bin">
    </use>
  </svg>
  <span class="u-sr-only">
    Delete
  </span>
</a>
```
dangerous

I guess they could be cool to print on a list view e.g.

### LINKED ISSUE #591: button generator
For the button variants change colors, shadows, etc, it’d be great to have a generator where we can specify the button name and the map of properties for each state. Some of the button variants can then be removed, and specified purely in a sass map.

Perhaps it also makes sense to generate the other button variants that change more than colors?

planktonic:
Fixes #591 
Fixes #573 

- Adds a button generator
- Uses it to create the base button colors & states, and various sizes

Delete if not applicable:

- [ ] Storybook documentation has been updated
- [ ] Your changes have been added to the `unreleased` section of [CHANGELOG.md](../CHANGELOG.md)


planktonic:
Closing this, as the approach is pretty heavy-handed. This could probably be done much more simply if the base button component exposed its settings as CSS variables. Further buttons can then re-define those. This would remove the need for a generator for use by consumers of bitstyles — the API exposed but the CSS variables would do exactly the same.

PR #614, opened 2022-02-01
Update workflow name and badges

klappradla:
- update workflow name
- add badge for CI to README
- update npm badge to a fancier one

Note: we may need to update which workflow is required for PRs as the name changed...

### DIFF
--- a/.github/workflows/run-checks.yml
+++ b/.github/workflows/run-checks.yml
@@ -1,7 +1,7 @@
-name: Run linting & tests
+name: Linting & tests
 on: [push]
 jobs:
-  lint:
+  lint-and-test:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout 🛎️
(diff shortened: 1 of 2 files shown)


planktonic (approved):
Can’t see the results until it’s merged I guess? But looks like some much-needed updates 🎉

klappradla:
> Can’t see the results until it’s merged I guess?

You can in the rich diff:

Screenshot 2022-02-01 at 11 55 02

planktonic:
@klappradla Aha, thanks :D  I’d say maybe it would read better as `Linting & tests | Passed`? (without “run”) But that’s a total nitpick, I’m happy with this as it is ⚡️

klappradla:
> @klappradla Aha, thanks :D I’d say maybe it would read better as `Linting & tests | Passed`? (without “run”) But that’s a total nitpick, I’m happy with this as it is ⚡️

It will be exactly that. That's what my name change in the actions file was for.

.github/workflows/run-checks.yml
```diff
@@ -1,7 +1,7 @@
-name: Run linting & tests
+name: Linting & tests
```
planktonic, line 1: > (only change after the merge is the name on the label 🤷 )

Aha, now I understand that sentence :+1: Looks great!

PR #628, opened 2022-02-09
Unify NodeJS setup

klappradla:
The asdf version manager is able to read `.nvmrc` files if one specifies
`lecacy_version_file = yes` in one's `.asdfrc` file.

We can expect people to know their version managers, so let's also skip
the detailed steps on how to install a version via `nvm` or `asdf`.

The version in the `.tool-versions` file was more recent, that's why I
opted for this in the `.nvmrc` file.


klappradla:
What version manager are you using @planktonic ? Do you prefer the `.nvmrc` file, or the `.tool-versions` file?

Also, I cannot run storybook (I thought it was due to the Node version, that's why I came up with this draft here):

```
ModuleBuildError: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
TypeError: this.getOptions is not a function
```

😢 

What exact Node version are you on?

planktonic:
@klappradla I’m on `v14.18.1`, and I use asdf these days. But ideally the setup is as tools-friendly/agnostic as possible. Totally fine with deleting the instructions for the npm install 👍

planktonic (commented):
I feel like it’d be better to keep the tools-versions file over the nvmrc, why did you choose to stick to the older one? 🤔 (it seems more flexible that it applies for any tool even if it’s not node, though I realise that’s probably not a likely issue. I’m happy to go with it, just curious!)

klappradla:
> I feel like it’d be better to keep the tools-versions file over the nvmrc, why did you choose to stick to the older one? 🤔 (it seems more flexible that it applies for any tool even if it’s not node, though I realise that’s probably not a likely issue. I’m happy to go with it, just curious!)

I went with `.nvmrc` because it's supported by multiple version management tools. But if `asdf` is the tool of choice for most people on the project, then let's switch this. Will update in a second.

klappradla:
> ...Will update in a second.

... when I find the time 😉

klappradla:
Sorry it took me so long to get back to this @planktonic 🙈 

I moved the change over to the new contribution guide.

planktonic (approved):
Thanks for this! ✨

PR #630, opened 2022-02-09
hotfix/update-storybook

planktonic:
Updates storybook, as the build was failing. That added the eslint-storybook plugin, and updated to better handle webpack v5 (which we need for the colors exports to work)


klappradla (approved):
🚒

PR #631, opened 2022-02-10
Update dependencies 2022-02-12

planktonic:

- Updates all dependencies with `yarn upgrade`
- Manually updated cssnano, autoprefixer, analyse-css, and the webpack loaders 
- storybook now uses postcss@v8
- Updates fixtures. Apparently the output order or properties has changed again (back to the previous order as far as I can tell), after a CSSNano update. I checked the output characteristics (using css-stats), it remained the same, and a visual check over the storybook looks correct.

there are some remaining vulnerabilities reported by `yarn audit`, they’re all in storybook so we’ll have to wait for an update there


.storybook/main.js
```diff
@@ -8,7 +8,14 @@ module.exports = {
   addons: [
     '@storybook/addon-links',
     '@storybook/addon-a11y',
-    '@storybook/addon-postcss',
+    {
+      name: '@storybook/addon-postcss',
+      options: {
+        postcssLoaderOptions: {
```
planktonic, line 14: storybook now uses PostCSS8

test/css-stats.json
```diff
@@ -1,135 +1,186 @@
 {
-  "generator": "analyze-css v0.13.0",
+  "generator": "analyze-css v2.1.18",
```
planktonic, line 2: Because of the updated analyse-css version, the results here are different. I checked this branch with the old version of css-analyse and got the same results as are shown on the left (the old results), which means the CSS is the same (also checked visually of course)

klappradla (approved):
Nice ✨

test/css-stats.json
```diff
@@ -1,135 +1,186 @@
...
     "duplicatedSelectors": [
       "html (3 times)",
-      "fieldset (2 times)",
-      "legend (2 times)",
-      "[type=checkbox], [type=radio] (2 times)"
+      "legend (2 times)"
     ]
   }
-}
+}
```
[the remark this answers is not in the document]
planktonic: Not that I can see, but I formatted the `json` files with prettier, and added them to the `checks 👍

planktonic:
@klappradla
> Wild! You know https://classic.yarnpkg.com/lang/en/docs/cli/upgrade-interactive/#toc-yarn-upgrade-interactive ?

oooo, nice! Thanks for the link 💖

PR #633, opened 2022-02-16
[#499] Add component: skip-to-content link

### LINKED ISSUE #499: Add component: skip-to-content link

planktonic:
Fixes #499

- prettier is now run in post-hooks, to allow the `--fix` parameters to be passed to lint
- skip-link positioning and appearance
- skip-link docs

## looks like (default settings)

Screenshot 2022-02-16 at 09 37 10

Delete if not applicable:


package.json
```diff
@@ -6,14 +6,17 @@
   "scripts": {
     "build": "sass scss/bitstyles.scss build/bitstyles.css --style=compressed --no-source-map",
     "postbuild": "postcss build/bitstyles.css --replace",
-    "lint:scss": "stylelint --ignore-path .gitignore 'scss/**/*.scss' --customSyntax postcss-scss && yarn prettier:scss",
-    "lint:js": "eslint --ignore-path .gitignore --ext js,jsx . && yarn prettier:js",
-    "lint:docs": "prettier --ignore-path .gitignore --check './**/*.{md,mdx}'",
-    "lint": "yarn lint:scss && yarn lint:js && yarn lint:docs",
+    "lint:scss": "stylelint --ignore-path .gitignore 'scss/**/*.scss' --customSyntax postcss-scss",
+    "postlint:scss": "yarn prettier:scss",
```
klappradla, line 10: Wow, neat! Did not know this existed 💡

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Two functions (`shadow.to-box-shadow()` and `shadow.to-drop-shadow()`) for translating those shadows (which are specified in Sass' list format) into either CSS `box-shadow` format, or CSS filter `drop-shadow` format. Not necessary for the shadows accessed using `shadows.get()`
 - New `u-shadow` and `u-drop-shadow` utility classes to set box-shadows and drop-shadows respectively
 - The smallscreen version of the Sidebar organism can now be positioned to the left or right side of the screen by overriding `$bitstyles-sidebar-small-location` (defaults to `left`)
+- New atom component `a-skip-link`, for skip-to-content links
```
klappradla, line 12: I did not know of this concept 💡 

Thanks! It's [this](https://www.w3.org/TR/2012/NOTE-WCAG20-TECHS-20120103/G1), right?

Why is it called `skip-link` and not `skip-to-content`? Is this the general term for it?

scss/bitstyles/atoms/skip-link/skip-link.stories.mdx
```diff
@@ -0,0 +1,31 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
+
+<Meta title="Atoms/Skip link" />
+
+# Skip link
+
+A skip-to-content link for keyboard users, especially those with litlle or no vision, enabling them to skip your navigation and get straight to the content. The default colors are black & white for readability, but they can be [customized](#customization).
```
klappradla, line 7:
```suggestion
A skip-to-content link for keyboard users, especially those with little or no vision, enabling them to skip your navigation and get straight to the content. The default colors are black & white for readability, but they can be [customized](#customization).
```

scss/bitstyles/atoms/skip-link/skip-link.stories.mdx
```diff
@@ -0,0 +1,31 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
+
+<Meta title="Atoms/Skip link" />
+
+# Skip link
```
klappradla, line 5: What do I have to do, to actually "see" this in Storybook? Maybe we should render this story differently 🤔

klappradla (commented):
Thanks for pointing me towards that 💡 Did not know the concept before.

I'm a bit unsure about the stories for the new atom: I don't see anything there 🔍

CHANGELOG.md
planktonic: Exactly — it’s  just shorter, it’s become the common term for it

scss/bitstyles/atoms/skip-link/skip-link.stories.mdx
planktonic: Aha, ok maybe we should add a note above the example — you need to focus the link, otherwise it won't show. That means clicking inside the story then hitting tab so the focus goes to the link

package.json
planktonic: It useful! But it seems to run in a new process, so I’m not 100% sure about it — results can output to stdout after the main task has run. wdyt, should we stick with this, or use npm-run-all or something similar?

klappradla: Oh! So it runs in the background and then pops up out of the nowhere once it's done? That's not really useful then though 😞 

It's generally recommended to run formatting **before** linting as they can easier cancel each other out the other way.

So maybe the setup is just easiest to control if we keep it as it was: separate task for `prettier` and `linters` and then additional tasks that combine them 🤷 

`npm-run all` would do things in parallel, right? Maybe a nice addition on top 🍨 But probably not suuuper necessary as most people have the tools integrated in their editor setup anyways and then just use the tasks as a pre-flight check.

planktonic: thanks for the feedback! I removed the `post` hooks, and now the `lint:scss` and `prettier:scss`(and for js & mdx) tasks are back. So that the `--fix` option works, prettier is no longer run in those individual lint tasks, but it is when running the all-in-one `yarn lint`. Does that work, wdyt?

CONTRIBUTING.md
```diff
@@ -34,6 +34,32 @@ This project uses [stylelint](https://github.com/stylelint/stylelint) and [prett
...
+```sh
+yarn lint:js
+yarn lint:scss
+yarn lint:docs
+```
+
+This is useful for automatically fixing linting errors by passing the `--fix` parameter (which will not work with the rolled-into-one `lint` command, only these individual commands) e.g.
+
+```sh
+yarn lint:scss --fix
```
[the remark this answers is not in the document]
planktonic: Yes it’s correct as it is :) Yarn doesn’t need the extra `--` to pass parameters into its scripts

PR #639, opened 2022-03-10
Hotfix: prettier the files

planktonic:
## Changes

- linting scripts now use `&&` again, so they fail properly
- runs prettier on the json file


ArinzeJeffrey-droid (approved):

ArinzeJeffrey-droid:
🚀

PR #647, opened 2022-04-06
[#641] Add utility: inset

### LINKED ISSUE #641: Add utility: inset
Utility to output top/right/bottom/left positions. Default to only `0` for each. Default to being available at no breakpoints, but configurable.

planktonic:
Fixes #641 

## Changes

- Adds an `inset` utility class that outputs top/right/bottom/left
- Updates the `get-classname` helper function to output property names too, and uses that instead of the inline interpolation in `properties.output-directional()`
- 🚓DRIVE-BY: fixes a couple of typos in the margin docs

## Checks

Delete if not applicable:

### DIFF
--- a/scss/bitstyles/utilities/inset/inset.stories.mdx
+++ b/scss/bitstyles/utilities/inset/inset.stories.mdx
@@ -3,3 +3,82 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
 <Meta title="Utilities/inset" />
 
 # Inset
+
+Specify the various directional inset properties (top/right/bottom/left) on an element. These are only useful on an element that has something other than `static` positioning — that means elements with position `relative`, `absolute`, `fixed`, and `sticky`. For elements with relative-, absolute-, or sticky-positioning, the inset will be relative to their nearest non-statically positioned parent elements. For those with position `fixed`, the inset will be relative to the viewport.
+
+Default configuration provides a value of `0` for each of the properties at the four cardinal directions, and is not available at other breakpoints. See [customization](#customization) below for details on how to change that.
+
+<Canvas>
+  <Story name="u-0-top" inline={false} height="12rem">
+    {`
+      <div class="u-bg-brand-2 u-padding-m" style="width: 20rem">
+        <div class="u-bg-background u-padding-m" style="height:10rem;">
+          <div style="position:absolute;" class="u-0-top u-padding-m u-bg-white u-drop-shadow-default">u-0-top</div>
+        </div>
+      </div>
+    `}
+  </Story>
+  <Story name="u-0-right" inline={false} height="12rem">
+    {`
+      <div class="u-bg-brand-2 u-padding-m" style="width: 20rem">
+        <div class="u-bg-background u-padding-m" style="height:10rem;">
+          <div style="position:absolute;" class="u-0-right u-padding-m u-bg-white u-drop-shadow-default">u-0-right</div>
+        </div>
+      </div>
+    `}
+  </Story>
+  <Story name="u-0-bottom" inline={false} height="12rem">
+    {`
+      <div class="u-bg-brand-2 u-padding-m" style="width: 20rem">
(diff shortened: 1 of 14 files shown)


scss/bitstyles/utilities/inset/inset.stories.mdx
```diff
@@ -0,0 +1,84 @@
...
+<Meta title="Utilities/inset" />
+
+# Inset
+
+Specify the various directional inset properties (top/right/bottom/left) on an element. These are only useful on an element that has something other than `static` positioning — that means elements with position `relative`, `absolute`, `fixed`, and `sticky`. For elements with relative-, absolute-, or sticky-positioning, the inset will be relative to their nearest non-statically positioned parent elements. For those with position `fixed`, the inset will be relative to the viewport.
+
+Default configuration provides a value of `0` for each of the properties at the four cardinal directions, and is not available at other breakpoints. See [customization](#customization) below for details on how to change that.
+
+<Canvas>
+  <Story name="u-0-top" inline={false} height="12rem">
```
iampeternguyen, line 12: this is kind of surprising actually, i would have assumed it would be `u-inset-0`?

planktonic: The `inset` property is too new to use in projects — support in browsers only started about a year ago. So these inset classes all specify direction. The config could be changed to add an `inset` value, but I don’t think that belongs in the default config due to lack of support, wdyt?

For the order of the direction & distance (`0`/`top`), yes that doesn’t read so nicely :/ It’s the naming convention that we used from the beginning (see `u-margin-m-left`, the distance comes first, then the direction), which I think we should probably change, but if so then that should be done in a single PR (and a major version release, it involves lots of breaking changes!)

Or was it something else about the name that you meant?

I created an issue for it, feedback welcome 🙏 https://github.com/bitcrowd/bitstyles/issues/655

iampeternguyen: yeah i meant the naming. is the current class style `u-0-top`? i guess i can get used to that but i would have thought it would have also included the inset property, `u-inset-0-top`

planktonic: I see 🤔 I went with just `0` and `top` because that’s what the CSS is (`top: 0`), and it’s short. There’s no use of the `inset` CSS property. I wouldn’t have used the name `inset` at all, except I needed a name for the group of classes that apply top/right/etc :D 

I was also imagining that in the future we would switch the order of all our classes to match the CSS convention more closely, so this class would become `u-top-0`. That seems to me the most readable & short classname.

But what’s the reason for including `inset` before the size and direction? Given that `inset` can be a value for this direction of this class, we could end up with `u-inset-0-inset`

scss/bitstyles/utilities/inset/inset.stories.mdx
```diff
@@ -0,0 +1,84 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
+
+<Meta title="Utilities/inset" />
```
klappradla, line 3: Was just wondering: now that we have a load of utilities, would it make sense to group them further? E.g. `Utilities/Position/inset`?

scss/bitstyles/utilities/inset/inset.stories.mdx
iampeternguyen: oops sorry never responded to this. yeah that looks good to me then!

scss/bitstyles/utilities/inset/inset.stories.mdx
planktonic: Yes good idea 🤔 There’s probably others that could be grouped too — issue here: https://github.com/bitcrowd/bitstyles/issues/659

planktonic:
@klappradla thanks for the feedback! Inset stories are now scrollbar free 🙏

(scrollbars are turned to always-on)
Screenshot 2022-04-28 at 16 12 51

PR #652, opened 2022-04-11
[#651] Generate line-height utility class

### LINKED ISSUE #651: Generate line-height utility class
Generate, and use the standard mechanisms for allowing configuration

planktonic:
Fixes #651

## Changes

- line-height utility class is now generated in the standard way

## Checks

Delete if not applicable:

### DIFF
--- a/scss/bitstyles/utilities/justify/justify.stories.mdx
+++ b/scss/bitstyles/utilities/justify/justify.stories.mdx
@@ -96,7 +96,7 @@ These classes are available at the `m` breakpoint.
 
 ## Customization
 
-Customize the available justify-content values by overriding the `$bitstyles-justify-values` Sass map. Provide the name to be used for the class as the key, and the favlue for the `justify-content` property:
+Customize the available justify-content values by overriding the `$bitstyles-justify-values` Sass map. Provide the name to be used for the class as the key, and the value for the `justify-content` property:
 
 ```scss
 @use '~bitstyles/scss/bitstyles/utilities/justify' with (
--- a/test/scss/fixtures/bitstyles-overrides.css
+++ b/test/scss/fixtures/bitstyles-overrides.css
@@ -1899,8 +1899,8 @@ table {
     justify-content: start;
   }
 }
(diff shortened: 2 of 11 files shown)


scss/bitstyles/utilities/justify/justify.stories.mdx
```diff
@@ -96,7 +96,7 @@ These classes are available at the `m` breakpoint.
 
 ## Customization
 
-Customize the available justify-content values by overriding the `$bitstyles-justify-values` Sass map. Provide the name to be used for the class as the key, and the favlue for the `justify-content` property:
+Customize the available justify-content values by overriding the `$bitstyles-justify-values` Sass map. Provide the name to be used for the class as the key, and the value for the `justify-content` property:
```
klappradla, line 99: 🧑‍🏫

test/scss/fixtures/bitstyles-overrides.css
```diff
@@ -1899,8 +1899,8 @@ table {
     justify-content: start;
   }
 }
-.bsu-line-height-min {
```
klappradla, line 1902: 👀 what was that?

klappradla (approved):
Nice - consistency for the win 🍰

test/scss/fixtures/bitstyles-overrides.css
planktonic: Which part? :D This was the old version of the utility class — it was hard coded to just `min` (which is the minimum line-height where fonts are guaranteed to show the diacritics like `ü` or `é` without cutting them off.

This is in the “overrides” fixture, where each module has a variable overridden before the CSS is output. One of those is the namespace `bs` which is prefixed to the rest of the classname. In this case it was being incorrectly concatenated, so there’s no minus: `.bsu-line-height-min` should be `.bs-u-line-height-min` (and it is now! 🎉)

klappradla: > … In this case it was being incorrectly concatenated

This part.

PR #664, opened 2022-06-08
Update dependencies 2022 06 08

planktonic:
## Changes

- yarn upgrade
- upgrade storybook
- Removes the `-ms` prefixed  `placeholder` & `user-select`  CSS, as it’s no longer required to match our browser support matrix


ArinzeJeffrey-droid (approved):
Looks good <3

PR #674, opened 2022-08-15
[#666] size naming

### LINKED ISSUE #666: Size naming
A couple of issues with our current naming:

- gap sizes and padding/margin sizes don’t match. This was a decision taken on the basis that a “large” grid gap is `xxxl`, but e.g. a large icon could be `xl` — the name of component’s size is relative to that component instead of to the global sizing scale. With the rise of utility class usage, that’s confusing tho, and means developers need to look up e.g. what size a grid gap is in order to match it with an icon, or padding on another element.
- this is probably a good time to switch to using `l`, `xl`, `2xl`, `3xl` etc. instead of the current `l`, `xl`, `xxl`, `xxxl`

ArinzeJeffrey-droid:
Fixes #666

## Changes

- Updates sizes scale naming to use the numbered format i.e `s,xs,2xl...`

## Checks

Delete if not applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - The color palette generator in `tools/color` is renamed from `color.generate-palette` to `color.generate-scale`, and now produces a more evenly-spread range of colors, with more accurate saturation and hue than before
 - By default there are now two tiers of color palettes — the one used in your codebase and in all bitstyles components (the “semantic palette”) is generated from a base palette. In the default configuration, this base palette is generated from the base colors provided. See [https://bitcrowd.github.io/bitstyles/?path=/docs/design-tokens-color-palette--page](https://bitcrowd.github.io/bitstyles/?path=/docs/design-tokens-color-palette--page) for details on customizing this.
 - Default naming of color variants has switched from using numbers, to using meaningful names like `light`, `dark`, `base`. If you override the palette in your codebase, you may need to override the colors used by other bitstyles components
+- Sizes scale naming has been updated to use to following convention `s`, `xs`, `2xs`, `3xs`, and `l`, `xl`, `2xl`, `3xl`. Any utility classnames in your codebase containing the sizes (e.g. `u-margin-xxs-bottom`, `u-padding-xxxl`) will need to be updated to use the new naming scheme (e.g. `u-margin-2xs-bottom`, `u-padding-3xl`). Any uses of `size.get()` in your codebase will need to be updated to use the new naming scheme.
 
 ### Fixed
 
--- a/scss/bitstyles/atoms/avatar/_settings.scss
+++ b/scss/bitstyles/atoms/avatar/_settings.scss
@@ -1,6 +1,6 @@
 @use '../../tools/size';
 
 $sizes: (
-  'm': size.get('m'),
-  'l': size.get('l'),
+  'm': size.get('xl'),
+  'l': size.get('xl') + size.get('m'),
 ) !default;
--- a/scss/bitstyles/atoms/avatar/_settings.scss
+++ b/scss/bitstyles/atoms/avatar/_settings.scss
@@ -1,6 +1,6 @@
 @use '../../tools/size';
 
 $sizes: (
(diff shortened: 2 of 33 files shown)


CHANGELOG.md
```diff
@@ -16,10 +16,8 @@
...
 - Newly configured `line-height` utility classes to include the base `typography.$line-heights`
 - The `output-responsive-font-sizes` mixin expects a map data structure which includes `font-size` and `line-height`. If you were using the mixin you would have to change the parameter passed.
-
-### Changed
-
 - We now have a two-tier typography design token system. A single Sass map `$bitstylestypography-font-sizes` is used to create the existing responsive typography system for `h1` – `h6` elements, and the corresponding `.u-h1` – `.u-h6` responsive typography utility classes. The responsive typography for the base heading elements and the corresponding utility classes is built on top of these base design tokens. To update, all your unique font-sizes should be defined in the base scale in `settings/typography`, then referenced using `font-size.get()` to build your responsive typography scale in `settings/typography-responsive`.
+- Sizes scale naming has been updated to use to following convention `s,xs,2xl,...`
```
planktonic, line 20: It’s helpful to include a short note here on how users of bitstyles should change their codebase to deal with the change. Something like `If you already overrode the sizes with your own naming scheme, you don’t need to do anything. If you’re using the default naming scheme, replace all instances of `xxxl` in your codebase with `3xl`, `xxl` with `2xl`, `xxxs` with `3xs`, and `xxl` with `2xs`, wdyt?

Let’s also make the scale shown in the first sentence correct: `3xs, 2xs, xs, s, m, l, xl, 2xl, 3xl`

scss/bitstyles/atoms/avatar/_settings.scss
```diff
@@ -1,6 +1,6 @@
 @use '../../tools/size';
 
 $sizes: (
-  'm': size.get('xl'),
-  'l': size.get('xl') + size.get('m'),
+  'm': size.get('m'),
```
planktonic, line 4: This makes a big difference to the size of these components! Let’s keep the changes to utility classes only please, no components (the components need to be updated only where they reference sizes that have had their names changed, like `xxxl` or `xxl`)

scss/bitstyles/utilities/gap/_settings.scss
```diff
@@ -1,9 +1,14 @@
...
+  '2xs': size.get('2xs'),
   'xs': size.get('xs'),
   's': size.get('s'),
-  'm': size.get('l'),
-  'l': size.get('xl'),
+  'm': size.get('m'),
+  'l': size.get('l'),
+  'xl': size.get('xl'),
+  '2xl': size.get('2xl'),
+  '3xl': size.get('3xl'),
```
planktonic, line 12: I wasn’t expecting to make all the sizes available as gap sizes 🤔 Would this have been useful in some past circumstances?

planktonic (commented):
Remember that all the classnames used in stories and throughout the MDX files should be updated too — `u-margin-xxxl` and its directional variants will have new names (also padding, gap…)

scss/bitstyles/utilities/gap/_settings.scss
```diff
@@ -3,7 +3,7 @@
 $sizes: (
   'xs': size.get('xs'),
   's': size.get('s'),
-  'm': size.get('l'),
-  'l': size.get('xl'),
+  'm': size.get('m'),
+  'l': size.get('l'),
```
planktonic, line 7: This means we lost the `xl` -sized gap, and gained the `m`-sized gap. Probably we can:

- drop `m`, since it’s not been used anywhere or considered useful for the default config so far (unless you think it should be?)
- add `xl`
- rename places `u-gap-l` was used to `u-gap-xl`
- rename places `u-gap-m` ws used to `u-gap-l`

Then we end up with the same grid gaps, but with the names now matching the names of sizes in the design tokens map, wdyt?

planktonic (approved):
@ArinzeJeffrey-droid thanks for the updates, looks great! I rebased, this is ready to go

PR #682, opened 2022-08-29
[#681] Prefix custom property names

### LINKED ISSUE #681: Prefix custom property names
We prefix our classnames with a namespace (not by default, only when configured), and a layer prefix (since #678). Let’s standardise the way we output CSS variables too, and allow automatic prefixing using the same mechanisms

planktonic:
Fixes #681 

## Changes

- Adds a helper function that returns CSS custom properties in a standardized format — namespace, layer-prefix, then names passed in, all separated with `-` characters
- Uses `custom-property-get()` to create and use the CSS variables in the badge component

## Checks

Delete if not applicable:


ArinzeJeffrey-droid (approved):
🚀

PR #692, opened 2022-09-13
[#691] Rename u-gap-m to u-gap-l

### LINKED ISSUE #691: Rename u-gap-m to u-gap-l
`.u-gap-m` is still used in some of our documentation, but it was renamed to `u-gap-l`

- rename all instances of `u-gap-m` to `u-gap-l`

planktonic:
Fixes #691

## Changes

Cleanup after #674 — renames all the uses of `u-gap-m` in our documentation to `u-gap-l`, to match the renamed classname

## Checks

Delete if not applicable:

### DIFF
--- a/scss/bitstyles/atoms/card/card.stories.mdx
+++ b/scss/bitstyles/atoms/card/card.stories.mdx
@@ -9,7 +9,7 @@ A content wrapper that sections and subtly highlights a chunk of content. For la
 <Canvas>
   <Story name="Card">
     {`
-    <div class="u-grid u-grid-cols-3@m u-gap-m">
+    <div class="u-grid u-grid-cols-3@m u-gap-l">
       <article class="a-card">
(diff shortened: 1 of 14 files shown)


andreasknoepfle:
I first had a mild heartattack and though, oh no we need to rename all the things in our projects as well. But then I saw that the original change was just fixing the stories file 🤣. Good change.

andreasknoepfle (approved):

planktonic:
> I first had a mild heartattack and though, oh no we need to rename all the things in our projects as well. But then I saw that the original change was just fixing the stories file 🤣. Good change.

@andreasknoepfle yeh this is only cleaning up the docs :D But that was necessary because the gap classes were updated in another PR (and these instances of the classname were missed). For updating — the old classnames can be kept by changing the config to how it was before. (but this change was actually something requested, so maybe we do want to make the change to the new naming in our projects…)

PR #703, opened 2022-10-13
[#696] Remove some component-level mixins

### LINKED ISSUE #696: Remove some component-level mixins
We have many component-level or layout-related mixins that are no longer used. For example `absolute-cover` is achieved either in CSS or with the `inset` utility class. absolute-center is no longer needed since we have better layout options in modern CSS.

- remove those mixins that are no longer useful. 
- move any remaining mixins to a `mixin` folder, and leave only design token accessors and other helper functions in `tools`

planktonic:
Fixes #696

## Changes

- Removes the following mixins: list-reset, absolute-center, absolute-cover, break-text, icon, list-inline, sr-only, and vertical-center
- The following mixins have been moved from `tools/` to `mixins/`: aspect-ratio, clearfix, link, object-cover, truncate.

## Checks

Delete if not applicable:


scss/bitstyles/generic/_reset.scss
```diff
@@ -22,6 +22,10 @@ select {
   margin: 0;
 }
 
+ul {
+  padding: 0;
+}
```
iampeternguyen, line 27: does `list-style-type` not matter anymore? and do we need this reset for ol's as well?

iampeternguyen (approved):
just a small question but looks good to me otherwise

scss/bitstyles/generic/_reset.scss
planktonic: list-style-type is still specified using a classname. This PR doesn’t really do much anything with that, though it does set the basis for the follow-up PR at #704 which changes the classname from being an atom-level component class to being a utility class

But yes, this should be applied to `ol`s too, thanks! I’ll make that change here, and update the other PR so there’s no duplication

PR #707, opened 2022-10-19
[#686] Html & body stretch

### LINKED ISSUE #686: html & body stretch
In many projects, we apply something like the following

```css
html,
body,
#__next,
#root  {
  height: 100%;
  min-height: stretch;
}
```

This ensures the html & body elements are both at least as tall as the viewport, even when content does not fill that space. There is often one extra id selector, for either a next.js or react-js project (extra selectors could be configurable?)

Right now we have

```css
html {
  height: 100%;
}

body {
  min-height: 100%;
}
```

Which doesn’t work when a framework inserts its own root element (the id selectors above), and doesn’t account for the changing viewport height on mobile browsers.

planktonic:
Fixes #686 

## Changes

- Updates the height of body to `stretch`, to ensure it respects the variable height of mobile browser viewports
- The list of elements that style is applied to is now configurable
- Adds docs with an example of the layout

Screenshot 2022-10-19 at 12 36 24

## Checks

Delete if not applicable:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@
 ### Added
 
 - New helper function `custom-property.get()` for generating names of CSS custom properties that respect the namespace and layer-prefix configuration.
-- Most design tokens are now output as CSS Custom Properties in a `:root` block. All are prefixed with `bs` in the default configuration (customize that using `setup.$custom-property-namespace`), with names matching their Sass variable e.g. `--bs-color-brand-1-base`, `--bs-color-brand-1-light`, `--bs-size-xs`, `--bs-size-s`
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- Use `setup.$viewport-elements` to define which element should be considered equal to the viewport in size. these elements’ heights will match the viewport, including any variable sizing on mobile browsers.
```
iampeternguyen, line 7: so this would only work if all of the ancestor elements also have these styles right? 

```
html
body
-- #random-container
-----#app-container
```

so if the dom looks something like that, and i only pass `#app-container` would this still work? if not, maybe we can clarify?

iampeternguyen (approved):
LGTM with a minor question 🚢

CHANGELOG.md
planktonic: the intention here is that a wrapper element from a framework (like `#root` from react) can be added to the list. If there are multiple wrappers for some reason, they’d each need to be added to the list — I’ll make sure that’s clear in the docs, thanks!

(btw it does work with a second, unstyled wrapper in safari, but not in chrome or firefox: https://codepen.io/darrencadwallader/pen/GRBJReZ)

PR #709, opened 2022-10-19
Update Dependencies 2022-10-19

planktonic:
## Changes

- Simple `yarn upgrade`

### DIFF
--- a/test/scss/fixtures/bitstyles-overrides.css
+++ b/test/scss/fixtures/bitstyles-overrides.css
@@ -599,12 +599,12 @@ h3,
 h4,
 h5,
 h6 {
-  text-rendering: optimizeLegibility;
   font-family: FakeFont, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto,
-    helvetica, arial, sans-serif, Apple Color Emoji, Segoe UI Emoji,
-    Segoe UI Symbol;
+  helvetica, arial, sans-serif, Apple Color Emoji, Segoe UI Emoji,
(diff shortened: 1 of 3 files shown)


test/scss/fixtures/bitstyles-overrides.css
```diff
@@ -599,12 +599,12 @@ h3,
 h4,
 h5,
 h6 {
-  text-rendering: optimizeLegibility;
   font-family: FakeFont, -apple-system, BlinkMacSystemFont, Segoe UI, Roboto,
-    helvetica, arial, sans-serif, Apple Color Emoji, Segoe UI Emoji,
-    Segoe UI Symbol;
+  helvetica, arial, sans-serif, Apple Color Emoji, Segoe UI Emoji,
+  Segoe UI Symbol;
```
klappradla, line 604: Does this come from the linter/formatter? Makes it harder to read in my opinion, but if they think it's a good idea - fine.

klappradla (approved):
Fine 🚜 

Why is all the `webkit` stuff gone? No longer relevant these days?

planktonic:
> Why is all the `webkit` stuff gone? No longer relevant these days?

@klappradla yes that’s because of an update to the caniuse-lite database — the browsers which needed those prefixes are now considered to tiny in terms of usage (we’re using [browserslist defaults](https://browserslist.dev/?q=ZGVmYXVsdHM%3D), and it looks like some webkit or blink versions were removed)

PR #751, opened 2023-03-22
[#749] Update badge components

### LINKED ISSUE #749: Update badge components
We have new colors for the badge variants.

Refactor the way they’re built, so that colors can be manually configured, instead of always taking a set color from the scale.
- On smaller screens use the small badge variant

planktonic:
Fixes #749 

## Changes

- Badges are now more configurable — each color variant has its color & background-color specified explicitly in configuration instead of automatically. Defaults to grayscale.
- Badges now have size variants (base and small in default config), and always display as small on small screens
- Moves stories for Badge component into a `.js` file, and includes them into the mdx file
- Adds a component file for Badge — outputs a simple HTML element, mimicing how a component in e.g. React, web-components would work.
- Reactivates the controls plugin for storybook

## 📸

Firefox_Screenshot_2023-04-27T12-24-00 785Z

## How to QA Visual changes

- `git checkout main` 
- `git pull` 
- `git checkout feature/749-update-badge-components` 
- `yarn`
- `yarn storybook`
- navigate to Badge component

Check:

- [ ] The appearance matches the zeplin designs
- [ ] The documentation for this component is correct, understandable, and up-to-date.

## How to QA code changes

- [ ] The documentation for this component is correct, understandable, and up-to-date.
- [ ] The component makes good use of CSS custom properties to simplify creating variants (or doesn’t have variants).
- [ ] Everything that should be a variable, is.

## Preflight checks

PR author to check, and delete if not applicable:


ed-eva (commented):
* All (at least most) Shadows have an opacity of 0.075, it should be 0.2.
    * Note: I don’t know how to check for the shadow so effectively in web inspector though.

* Small badge-buttons are same size as regular badge-buttons (designs from my side needed here?)

* Badge-Buttons Active state - the icon should be white. Currently it is a light version of whatever colour.

* Colour of green text on “positive” badges is wrong - it should be (0, 153, 94) - it is currently (0, 153, 145) 
    * I think the colour of Brand-secondary’s text is being used here by mistake.

planktonic:
@ed-eva 

Thanks! I’ll work through this & update 👍 

> All (at least most) Shadows have an opacity of 0.075, it should be 0.2

Yes, this is related to the way we build shadows — default CSS box-shadows are flat and basic-looking, so we use multiple shadows on top of each other to build something a little prettier, a little more natural-looking. The result is that each layer of shadow needs to be a lower opacity, so they add up to a similar opacity all together.

> Small badge-buttons are same size as regular badge-buttons (designs from my side needed here?)

Yes exactly – right now we have no design for these :) I updated the ticket for badges: https://bitcrowd.atlassian.net/browse/ID-158

ed-eva (commented):
All Badges
* Shadow colour looks a bit intense compared to the designs, can we reduce opacity a tad?
* Warning
    * Text colour needs to be darker “Warning/Dark-2”
* Positive
    * Text colour needs to be darker “Positive/Dark-2”
* Secondary
    * Text colour needs to be darker “Secondary/Dark-2”

Badge-Buttons
* All
    * Maybe using a cross as the icon would be clearer - in the documentation we talk about them being removable but they have plus icons.
* All (except Danger)
    * States of the button need to be shifted to increase contrast. Hover is Dark-3 icon on Light-2 background. Active is a white icon on a Dark-2 background
* Danger
    * Hover is Dark-2 icon on Light-2 background & Active is white on Dark-1 background.

ed-eva (commented):
Looks good

scss/bitstyles/atoms/badge/badge.stories.mdx
```diff
@@ -1,117 +1,133 @@
...
 
-By default `gray`, `brand-1`, `brand-2`, and `danger` colors are available (see [Customization](#customization) section below). Here, they’re displayed in a list (there are often multiple tags or filters so this is very common).
+Size variants:
+
+- base (default)
+- small
+
+All badges are displayed as `small` badges on small screens.
+
+## “Extra” area
```
bluemoonecho, line 49: What about "prepend" for things on the left and "append" for things on the right?

scss/bitstyles/atoms/badge/badge.stories.mdx
```diff
@@ -1,117 +1,133 @@
...
+    ),
+  )
+);
 ```
+
+The size-related properties available for configuration are:
+
+| Property name        | Description                                                                                 |
+| -------------------- | ------------------------------------------------------------------------------------------- |
+| `padding-vertical`   | `padding-top` and `padding-bottom` properties                                               |
```
bluemoonecho, line 130: Any reasons why the words "vertical" and "horizontal" were preferred to the more traditional "x" and "y"?

bluemoonecho (approved):
Just left a couple of questions, but LGTM!

scss/bitstyles/atoms/badge/badge.stories.mdx
planktonic: Yeh that’s a great point 🤔 

`x`/`y` are used for the padding & margin utility classes, but the sass variables for components have always been named with `horizontal`/`vertical`. I think there’s no good reason for that difference though, and actually I’ve been wondering when it might make sense to switch to using `padding-block`, `margin-block`, `padding-inline` etc for these. Caniuse shows two years of support for these logical directions (https://caniuse.com/css-logical-props)

I’ll keep this naming for now, and add an issue to investigate switching to using the logical properties (and the matching naming) everywhere

Issue #782, opened 2023-05-23
Update Page Header component


ed-eva:
- The Page header component has been created using sections to mimic how it might be implemented, there are top, middle & bottom sections each split into left and right. 
- The section widths are often split 50/50, however the right middle section is often 'fill-width' so that components like badges arrange themselves next to the title. 

- There is a drop shadow on the bottom edge.
- On smaller screens items stack in order from left to right and top to bottom. (like western reading pattern)

Zeplin Link: https://zpl.io/qyBk38Z

PR #793, opened 2023-07-20
Task/internal 28 update page header component 782

bluemoonecho:
Issue #782

Zeplin link:  https://zpl.io/qyBk38Z

## 📸 Looks like

| 📱 | 🖥️ |
| --- | --- | 
| Screenshot 2023-08-01 at 15 10 23 | Screenshot 2023-08-01 at 14 28 11


planktonic:
@bluemoonecho yes the designs in zeplin should be complete, and that’s our target appearance for every component :) If there’s anything that seems out of place or somehow inconsistent in the designs though, shout out so we can update them.

The layout improvements here look good, I’d add the following points (some of these apply to all the upcoming tickets — perhaps we can get together to do some ticket grooming, to make them clearer, wdyt?)

- We’ll eventually upgrade to storybook 7, where we need all the html stories taken out of mdx files, and instead specify them in js (then import them into the mdx to render them in the docs). Let’s make that change here already
- At the same time, I’d suggest adding a `PageHeader` js component to use in the stories, like the `Badge` or `Button` components we already have.
- The page-header component would be a great candidate to make into a CSS component (e.g. `.o-page-header`) so that we do not need all these utility classes in the template
- The zeplin designs show several slots inside the page-header (in zeplin these are named `top-left`, `bottom-right` etc, but probably it would be good to use semantic names here, to avoid them being based on visual layout). Perhaps make each of those a prop on the js component, that accepts some HTML and renders it in place
- Let’s use CSS grid to create the layout, using named grid areas for each of the slots

That’s a lot, and I’m sorry it wasn’t on the ticket in the first place! Does it make sense? Give me a shout so we can take a look together — this will be the most complex component we’ve built so far using the new format, so there’ll be some experimentation involved

.stylelintrc.json
```diff
@@ -6,6 +6,10 @@
       {
         "ignore": "local"
       }
-    ]
+    ],
+    "function-whitespace-after": null,
+    "media-feature-range-operator-space-after" : null,
+    "media-feature-range-operator-space-before" : null,
+    "string-quotes": null
```
bluemoonecho, line 13: I believe `stylelint-config-bitcrowd` might be outdated. 
I added these changes to fix this lint error: 

Screenshot 2023-08-01 at 14 12 36

planktonic: Yes it needs to be updated for v15 of stylelint :+1: I added an issue over on our linting repo: https://github.com/bitcrowd/frontend/issues/39

scss/bitstyles/organisms/page-header/PageHeader.js
```diff
@@ -0,0 +1,156 @@
...
+    { position: 'top-left', content: topLeft },
+    { position: 'top-right', content: topRight },
+    { position: 'center-left', content: centerLeft },
+    { position: 'center-right', content: centerRight },
+    { position: 'bottom-left', content: bottomLeft },
+    { position: 'bottom-right', content: bottomRight },
+  ];
+
+  contentPositions.forEach((item) => {
+    if (item) {
```
planktonic, line 29: Is there ever a case where there’s no item? (I think `forEach` wouldn’t iterate over e.g. en empty array). Maybe this could become the check for item.content, so we avoid empty divs?

```suggestion
    if (item.content) {
```

scss/bitstyles/organisms/page-header/_index.scss
```diff
@@ -0,0 +1,66 @@
+@use '../../tools/media-query';
+@use '../../tools/classname';
+@use '../../tools/design-token';
+
+#{classname.get($classname-items: 'page-header', $layer: 'organism')} {
+  padding-top: var(design-token.get('size', 'm'));
+  background-color: var(design-token.get('color', 'grayscale', 'light-3'));
+}
+
+.o-page-header__content {
```
planktonic, line 10: Each of these classnames also needs to be generated, same as in line 5 — perhaps instead, this is an ideal place to make use of sass nested classnames? e.g.

```scss
&__content {}
```
(and nested under the root classname above)

scss/bitstyles/organisms/page-header/page-header.stories.js
```diff
@@ -0,0 +1,56 @@
...
+    },
+    centerRight: {
+      description: 'The top left area of the header',
+    },
+    bottomLeft: {
+      description: 'The top left area of the header',
+    },
+    bottomRight: {
+      description: 'The top left area of the header',
+    },
```
planktonic, line 24:
```suggestion
    centerRight: {
      description: 'The center right area of the header',
    },
    bottomLeft: {
      description: 'The bottom left area of the header',
    },
    bottomRight: {
      description: 'The bottom right area of the header',
    },
```

scss/bitstyles/organisms/page-header/_index.scss
```diff
@@ -0,0 +1,66 @@
...
+  @include media-query.get('m') {
```
planktonic, line 11: Can we group all the media-query styles into one block at the end of the Sass (instead of several blocks, each inside their page-header__content-section class)? That way, the CSS that’s generated is clearer and smaller

scss/bitstyles/organisms/page-header/_index.scss
```diff
@@ -0,0 +1,66 @@
...
+    display: grid;
+    grid-template-columns: repeat(2, minmax(0, 1fr));
+  }
+}
+
+.o-page-header__top-left {
+  @include media-query.get('m') {
+    grid-column: 1;
```
planktonic, line 19: Using `grid-template-area` may make for more readable CSS here, wdyt? (https://css-tricks.com/snippets/css/complete-guide-grid/#aa-grid-template-areas)

bluemoonecho:
@planktonic - should be ready for another review!

planktonic (approved):
Looks great! I left a few suggestions, and it’s good to go

Issue #763, opened 2023-04-12
Update Sidebar component


planktonic:
- Uses the invisible button variant 
- Buttons fill the container width 
  - Have a padding of 0.5rem to the sides. 
  - Have a spacing of 0.75rem between buttons 
- A divider of 2px can be used at the bottom if showing an account avatar or CTA. 
- There is a dropdown shadow.
- When there are many buttons the sidebar scrolls
  - The bottom section, logo and close icon remain sticky when scrolling.

Zeplin link: https://zpl.io/5EKZ8JW

PR #794, opened 2023-07-24
Task/internal 31 update sidebar component 763

bluemoonecho:
Takes care of issue #763

[Zeplin](https://zpl.io/5EKZ8JW)
[Ticker](https://bitcrowd.atlassian.net/browse/INTERNAL-31)

## Changes

- [x] Uses the invisible button variant
    - [x] Buttons fill the container width
    - [x] Have a padding of 0.5rem to the sides.
- [x] Have a spacing of 0.75rem between buttons
- [x] A divider of 2px can be used at the bottom if showing an account avatar or CTA.
- [x] There is a dropdown shadow (only on mobile)
- [x] When there are many buttons the sidebar scrolls
- [x] The bottom section, logo and close icon remain sticky when scrolling.

## 📸 Looks like

| 📱 | 🖥️ |
| --- | --- |
|  Screenshot 2023-07-24 at 16 51 46 | Screenshot 2023-07-24 at 16 40 24


scss/bitstyles/ui/sidebar.stories.mdx
```diff
@@ -81,73 +81,74 @@ Uses a fullwidth dropdown to fit with no overlap.
     {`
       <div class="u-flex u-height-100vh" x-data="{ sidebarOpen: false }">
```
planktonic, line 82: Can we build a sidebar component here, which accepts the content slots (that we want to provide different content to, so perhaps the logo can be hardcoded) as props? Then this story could be moved to a JS file, and imported into the mdx (as we transition to storybook 7, we’ll need to do this)

scss/bitstyles/ui/sidebar.stories.mdx
```diff
@@ -81,73 +81,74 @@ Uses a fullwidth dropdown to fit with no overlap.
...
                   </svg>
                   Dashboard
                 </a>
               </li>
-              <li class="u-margin-s4-bottom u-flex">
-                <a href="/" class="a-button a-button--transparent u-flex-grow-1" aria-current="page">
-                  <svg width="16" height="16" class="a-button-nav__icon a-icon a-icon--m u-margin-s2-right u-margin-neg-s6-left" aria-hidden="true" focusable="false">
-                    <use xlink:href="${icons}#icon-filter"></use>
+              <li class="u-margin-s2-bottom u-flex">
+                <a href="/" class="a-button a-button--transparent u-flex-grow-1 u-justify-start" aria-current="page">
```
planktonic, line 96: For all of these tickets, where we have a component that we‘ve built in JS (e.g. `Badge` and `Button`), let’s use those components in places like this — for consistency & maintenance, all the usual component advantages :)

planktonic (commented):
The sidebar looks great! ✨

The buttons look good too, but they’re different to the zeplin designs — is there some inconsistency there, do we need to update the designs?

A final note: can you check that the zeplin components are linked to the stories please, so we can more easily QA these against designs? 🙏

scss/bitstyles/ui/sidebar.stories.mdx
```diff
@@ -3,22 +3,14 @@ import logoLarge from '../../../test/assets/images/logoLarge.svg';
 import logoSmall from '../../../test/assets/images/logoSmall.svg';
 import icons from '../../../assets/images/icons.svg';
 
-<Meta title="UI/Navigation/Sidebar" />
+<Meta title="Organisms/Sidebar" />
```
planktonic, line 6: This has the same name as the `.js` file with the sidebar stories in, so neither of them shows up in storybook’s navigation (names need to be unique so storybook knows which is meant). Let’s keep this as being in the UI layer, since it shows complete examples on how to use this component
```suggestion
<Meta title="UI/Navigation/Sidebar" />
```

planktonic (approved):
One suggestion, so that storybook works properly again :) Then this is good to go! ⚡️

planktonic:
@bluemoonecho ah, tests are failing because the fixtures need updating — give me a shout if I can help get that sorted out

PR #798, opened 2023-08-01
Driveby - Fix CI node version.

bluemoonecho:
New changes on the github actions workflow. We now need to specify the node version. [See](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions#create-an-example-workflow)


planktonic (approved):
✨

PR #822, opened 2024-06-19
[#819] Insufficient 5.0.0 changelog entries

### LINKED ISSUE #819: Insufficient 5.0.0 changelog entries
While doing an upgrade from 4.3.0 to 5.0.0, I was surprised by a few changes. After reading the changelog, I was not expecting those changes. Here's a list.

Note whenever I refer to the changelog for 5.0.0, I am also including the changelog for 5.0.0-alpha.

## 1. `a-badge--gray`

The class `a-badge--gray` does not exist anymore. There is no mention of any changes to badges in 5.0.0. I would need to know how to replace that class and whether any other badge-related classes changed.

| before | after |
| --- | --- |
|    Screenshot 2024-04-19 at 12 22 41 |  Screenshot 2024-04-19 at 12 22 34  |

## 2. `u-fg-gray-30` and others

The classes `u-fg-gray-30` and `u-fg-gray-50` do not exist anymore. There is no mention of any color-related **class names** changing. In fact, the changelog suggests that I only need to react if I am overwriting default colors in any way:

> Default naming of color variants has switched from using numbers, to using meaningful names like light, dark, base. If you override the palette in your codebase, you may need to override the colors used by other bitstyles components

I would need to know all color-related class name changes because we need to remap them all in phoenix_bitstyles.

## 3. `u-gap-m`

The class `u-gap-m` does not exist anymore. All the other gap classes exist (`u-gap-s`, `u-gap-l` etc), only this one is missing. Is this a bug?

planktonic:
Fixes #819 

## Changes

Updates & adds changelog notes:

- about the badge component in the changelog, regarding changing variants
- about the gap utility class changing naming scheme (and therefore classnames) in the default configuration
- about handling the new color palette in fg and bg modules

## How to QA

Check that the notes cover every point mentioned in the issue raised, and that they accurately reflect the state of things at v5.0


angelikatyborska:
## RE: badge variant

All clear ✅ 

## RE: `.u-gap-m` missing

The addition to the changelog doesn't really tell me why `.u-gap-m` is missing. Additionally, I would need to know exactly how to translate **all** of the old `u-gap-XXX` classes to the new `u-gap-YYY` classes so that they have the same gap values. Something like:

```
u-gap-xs -> u-gap-???
u-gap-s -> u-gap-??
u-gap-m -> u-gap-???
and so on...
```

## RE: color palette

I would need to know exactly how to map **all** `u-fg-XXX` and `u-bg-XXX` utility class that existed in 4.3.0 to the new classes in 5.0.0. So not only `u-fg-gray-30`, but every other gray, and every other color that existed in the default config in 4.3.0 (I don't know exactly which colors those were)

```
u-fg-gray-10 -> u-fg-???
u-fg-gray-20 -> u-fg-???
u-fg-gray-30 -> u-fg-text-light
u-fg-gray-40 -> u-fg-text-light
u-fg-brand-1-10 -> u-fg-???
u-fg-brand-1-20 -> u-fg-???
and so on...
```

angelikatyborska:
Ok, maybe I am exaggerating with those two points:

> I would need to know exactly how to translate all of the old u-gap-XXX classes to the new u-gap-YYY classes 

> I would need to know exactly how to map all u-fg-XXX and u-bg-XXX utility class 


Maybe I don't need to know that, it should be enough to translate the classes that bitstyles phoenix used and only those classes 🤔 

From [this commit](https://github.com/bitcrowd/bitstyles/commit/3a74bc034ba5000b63d7d1694b55b8b4add30858) I conclude that `u-gap-m` is `now u-gap-l`

planktonic:
> The addition to the changelog doesn't really tell me why .u-gap-m is missing.

Generally the utility classes aren’t output for every design token on a scale. So the `fg` & `bg` classes don’t exist for every color, and the gap classes don’t exist for every size. The purpose is to provide some reasonable defaults, instead of the kitchen sink approach. And since there was no `.u-gap` class that output `gap: var(--bs-size-m)` before, there isn’t one now either.

Does that need to be explained in the changelog entry? So far, we’ve just been trying to describe the “how”, not so much the details of “why”. If you feel it it does, could you perhaps suggest something that would cover the gap in the explanation for you?

> I would need to know exactly how to translate all of the old u-gap-XXX classes to the new u-gap-YYY

Yes totally, I should’ve done that 👍  Updated!

planktonic:
> I would need to know exactly how to map all u-fg-XXX and u-bg-XXX utility class that existed in 4.3.0 to the new classes in 5.0.0. 

I’ve added a suggested list to the changelog entry, with the caveat that this cannot be an automatic migration — this needs a human to check the UI is still visible and readable, since the color utility classes are now a smaller set.

angelikatyborska (approved):
Perfect, thank you 🙏

PR #823, opened 2024-07-08
Allow users of bitstyles to run node 22

angelikatyborska:
This library delivers CSS only and doesn't need to require its users to run a specific version of node.

This change should remove the warning you would see when using node 22 and installing this library:
```
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'bitstyles@6.1.1',
npm warn EBADENGINE   required: { node: '^18.16.0', npm: '^9.5.1' },
npm warn EBADENGINE   current: { node: 'v22.4.0', npm: '10.8.1' }
npm warn EBADENGINE }
```


planktonic (approved):
Thanks!

